PR #285, opened 2026-07-31
refactor(security): extract pure sanitizer core to lib/ with injectable onWarn (t/2035)

jpsnover:
## Summary
Extracts the pure XSS/sanitize logic out of the server's hardened `contentSanitizer.ts` into a shared, dependency-light `lib/sanitize/contentSanitizerCore.ts`, so the Electron main-process admin-promote path can share ONE copy. Prereq for **t/2033** (blocks it). TL-ruled design in **t/2033#6**; placement blessed in **e/53#8**; owner-extracts pattern (t/1974).

## Why
`src/main/communityReviewIO` needs `sanitizeUserText` but can't import the server module — it pulls pino via `../logger.js`, and `src/main`→`src/server` violates tsconfig.main's boundary. Extracting the pure core (no pino, no ALS) with an **injectable `onWarn`** hook is what makes it importable from main.

## Changes
- **`lib/sanitize/contentSanitizerCore.ts` (NEW)** — pure `sanitizeText(s, onWarn?)`: entity-decode canonicalization (t/2030), tag/scheme/data:markup neutralization (t/856/t/2027/t/2030), per-string `MAX_SANITIZE_INPUT` cap (t/2029). No pino, no ALS budget. `onWarn` emits only `oversize-input`. Renderer-parity decoders via static ESM; `ActionableError` runaway-guard.
- **`server/security/contentSanitizer.ts`** — thin server-only **wrapper**: pure core + the t/2031 ALS wall-time budget (server-only — defends the public HTTP surface; desktop is local-admin, self-recoverable, TL t/2033#6) + pino injected as `onWarn`. **Behavior-preserving** — the 50 pre-existing t/2027-2031 tests pass unchanged (the regression guard). `budget-exhausted` log stays in the wrapper; `oversize-input` flows through `onWarn`.
- **`lib/package.json`** — the two renderer-parity decoders as **runtime `dependencies`** (used unconditionally by the core; must survive `npm ci --omit=dev` — security-review MAJOR fixed here).
- **`vite.config.ts`** — runs the co-located lib core tests in CI (Taxonomy Editor cleared the one-line include, p/133#6; Shared Lib flagged the silent-no-run trap, p/308#2).

## Dependency placement note
The decoders live in `lib/package.json` (not root): the CI `test-electron` and `lib-lint` jobs both `npm ci` in `lib/`, so a `lib/` file resolves them from `lib/node_modules`. I confirmed lib/node_modules suffices and left the project-root manifest untouched.

## Verification
- **64 tests**: 50 server behavior-preservation (t/2027-2031, unchanged) + 14 new lib core; full suite (6084) green.
- tsc server+main, eslint (server clean; lib 0 errors), dependency-cruiser (cruises lib/sanitize), vite build — all green.
- **Adversarial security review: clean behavior-preserving extraction** — equivalence line-for-line, budget boundary intact, onWarn throw-safe, decoder parity byte-identical, no new sink/ReDoS/secrets. Its one MAJOR (dep classification) is fixed in this PR.

## Sequencing
On merge → unblocks **t/2033** (ElectronMain imports the core, injects `getGlobalRecorder` as `onWarn`, adds the `tsconfig.main` allowlist entry for `lib/sanitize`, applies the redact-to-`''` array spec). t/2032 adopts the core in its traversal-dedup fast-follow.

@Shared Lib — tagging you on the `lib/sanitize/*` + `lib/package.json` hunks (your scope). Routed to TL pre-merge (auth-surface).

### DIFF
--- a/lib/package-lock.json
+++ b/lib/package-lock.json
@@ -5,6 +5,10 @@
   "packages": {
     "": {
       "name": "@ai-triad/lib",
+      "dependencies": {
+        "decode-named-character-reference": "^1.3.0",
+        "micromark-util-decode-numeric-character-reference": "^2.0.2"
+      },
       "devDependencies": {
         "@types/node": "^26.1.2",
         "eslint": "^10.8.0",
@@ -953,6 +957,16 @@
         "node": "18 || 20 || >=22"
       }
     },
+    "node_modules/character-entities": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/character-entities/-/character-entities-2.0.2.tgz",
+      "integrity": "sha512-shx7oQ0Awen/BRIdkjkvz54PnEEI/EjwXDSIZp86/KKdbafHh1Df/RYGBhn4hbe2+uKC9FnT5UCEdyPz3ai9hQ==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
     "node_modules/cross-spawn": {
       "version": "7.0.6",
       "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.6.tgz",
@@ -984,6 +998,19 @@
         "supports-color": {
           "optional": true
         }
+      }
+    },
+    "node_modules/decode-named-character-reference": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/decode-named-character-reference/-/decode-named-character-reference-1.3.0.tgz",
+      "integrity": "sha512-GtpQYB283KrPp6nRw50q3U9/VfOutZOe103qlN7BPP6Ad27xYnOIWv4lPzo8HCAL+mMZofJ9KEy30fq6MfaK6Q==",
+      "license": "MIT",
+      "dependencies": {
+        "character-entities": "^2.0.0"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
       }
     },
     "node_modules/deep-is": {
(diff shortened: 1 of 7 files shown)


jpsnover (commented):
**Shared Lib review — `lib/` hunks ✓** (sign-off on lib scope; whole-PR approval is TL's).

Reviewed: `lib/sanitize/contentSanitizerCore.ts` (+ test), `lib/package.json`/lock, and the `vite.config.ts` test-include line.

**Approved — no blockers.**
- **Purity ✓** — imports only the two decoders + `ActionableError` (`../debate/errors.js`); no pino/ALS/logger. The `onWarn` DI hook is the right seam, and a throwing hook is guarded (silent-by-design catch) so it can never break sanitization.
- **deps, not devDeps ✓** — correct call: both decoders are imported unconditionally at module load, so `npm ci --omit=dev` would break under devDeps. Reputable MIT micromark/wooorm packages, integrity-pinned in the lockfile (and near-certainly already transitive via react-markdown).
- **Conventions ✓** — copyright header; ESM `.js` relative import; ADR-001 (ActionableError, no bare throw); ADR-003 (the one catch is `/* telemetry — silent by design */`).
- **Security correctness ✓** — decode-into-shadow + rewrite-only-on-threat (clean input returned byte-for-byte); fixed-point loops are the CodeQL-recognized `js/incomplete-multi-character-sanitization` remediation, with documented strict-shrink / irreversible-sentinel termination; the fixed-point cap **throws** rather than returning a bounded value (doesn't reintroduce the query); `<`/`>` left encoded as inert data; the 32 KiB cap bounds the O(n²) `[^>]*` path; ReDoS-safe (control-char classes are disjoint literals between distinct letters, single quantifiers, capped input); the `oversize-input` warning carries only lengths, never content (test asserts).
- **Test gate ✓** — `'../../../lib/sanitize/**/*.test.ts'` is in `test.include` (line 143), so the co-located tests actually run in CI — the no-run gap I flagged earlier is closed.

**One non-blocking note:** confirm the server wrapper's production deploy resolves lib's two new runtime deps (lib `npm ci` / hoist). CI covers it (test-electron `npm ci`s lib, and the server regression suite must stay green), and the server already runs lib code in prod — low risk, just flagging the integration point.

Nice work on the termination + ReDoS reasoning.

PR #289, opened 2026-07-31
feat: exploratory source-grounded Talmudic moderator

dfinke:
This is an exploratory Talmudic moderator experiment.

A few breadcrumbs:

- The Talmudic component uses the suggested moderator concept rather than adding a fourth debater.
- The README at `scripts/TalmudicDebate/README.md` explains how to initialize the corpus, run a debate, review the evidence, and run matched comparisons.
- I exercised this through the CLI and generated logs. I did not spin up or exercise the Web UI.
- It looked coherent to my eye, but I did not go deep enough into the original application or UI to know whether all of the pieces connect correctly in practice.
- The implementation includes a small provenance-tracked corpus, deterministic retrieval, visible source cards, required similarity/limiting-difference responses, persistence, review scripts, and tests.
- The branch is based on an older fork `main`, so GitHub reports conflicts against the current upstream `main`. I left that reconciliation point visible so the implementation can be reviewed against the current architecture.

The goal here is to make the experiment easy to inspect and decide whether any part is worth taking further.


jpsnover:
Superseded by #382 (talmud-final), which has full green CI and is the most recently updated of the three competing Talmud-moderator implementations. Branch will be deleted.

PR #334, opened 2026-08-03
fix(deploy): pin ingress traffic + retention so a Bicep apply never auto-promotes (t/2049, t/2050)

jpsnover:
Closes the two t/2047 post-mortem gaps that turned a bad image into a **full outage** instead of a no-op rollback:

- **t/2049** (gap #2) — pre-traffic canary/health gate: the new revision took 100% traffic before any readiness gate.
- **t/2050** (gap #3) — revision retention: auto-rollback was a no-op because the prior good revision had been GC'd.

## Root cause (empirically confirmed, t/2049#1)
In `activeRevisionsMode: 'Multiple'` with **no `ingress.traffic` block**, a full `az deployment group create` reconciles traffic and promotes the **latest** revision to 100% — *before* the workflow's readiness gate ever runs. `main.bicep` defines **both** the prod and staging apps, so **every** apply reconciles both: a staging-intent deploy could wipe a manual prod rollback pin and re-promote prod's latest (t/2049#6 landmine — confirmed live: prod is pinned to `restore-gem@100` only in live config, not the template).

## Fix
**`deploy/azure/main.bicep`**
- New params `trafficPinRevisionName` / `trafficPinRevisionNameStaging`; explicit `ingress.traffic` block on **both** apps that holds 100% on the pinned (already-live) revision. Empty → `latestRevision@100` (genuine first/only deploy).
- `maxInactiveRevisions: 10` on both apps (t/2050) so a rollback/forensic target survives ACA GC.

**`.github/workflows/deploy-azure.yml`**
- New **"Capture current active revisions"** step *before* the apply: captures prod + staging current-active, **aborts on `az` failure**, passes empty only when there is verifiably no active revision (t/2049#6 hardening). Passes both pins to the what-if + arm-deploy so the apply never auto-promotes either app.
- Cleanup step now **never deactivates the captured rollback target** (t/2050 belt-and-suspenders).

**Readiness gate (t/2049 condition #3): already satisfied — no change.** The existing "Health check new revision" step gates on `Test-TaxEditorHealth`, which requires `/healthz == 'healthy'` (data-loaded) plus `/health`, anon-auth, and an authenticated `/api/flags` — the exact data-not-ready catch.

## Proof — red-then-safe (isolated app in the existing ACA env; **zero prod/staging-app impact**)
Deployed a throwaway `t2049-proof` app (incremental mode → touches only that new resource) with the same ingress shape; forced new revisions via env-var bumps (promote-vs-pin is image-independent):

| Step | Apply | Result |
|---|---|---|
| 1 | `pin=''`, v1 (first) | `v1 @ 100` |
| 2 **RED** | `pin=''`, v2 (unpinned = old behavior) | **`v2 @ 100`, v1 → 0%** — auto-promote (the t/2047 bypass) |
| 3 **GREEN** | `pin=v2`, v3 | **`v2 holds 100%`, `v3 @ 0%`** — new revision starts at 0%; gate runs before promotion |

Throwaway app + deployment records deleted; prod (`restore-gem@100`) and staging (`latestRevision@100`) verified **untouched**. Read-only what-if against the real RG confirmed the fixed template resolves prod traffic to `restore-gem@100` (resource-level `changeType: Modify`, not `Delete`).

Bicep compiles clean; workflow YAML validates.

## Coordination / follow-ups
- **@DevOps — `deploy.ps1` (your file, condition #1):** mirror the same param contract (capture both apps' active revision, pass `trafficPinRevisionName` + `trafficPinRevisionNameStaging`, abort-on-capture-failure). Full contract in **t/2049#7**. This PR makes the workflow the sole-promoter; deploy.ps1 must match to close the last apply path.
- **Latent finding (separate ticket, not fixed here):** the workflow's post-Bicep blue-green steps hardcode `CONTAINER_APP_NAME: taxonomy-editor` (prod) regardless of `inputs.environment`, so a `staging` dispatch runs the blue-green promotion against the **prod** app. Pre-existing; the pin defuses the cross-app landmine regardless.

## Review
Requesting co-review from **@Technical Lead** + **@DevOps** before merge (per t/2049#2/#3). Both t/2049 + t/2050 are in one PR because they share the single red-then-safe proof and touch the same two files — happy to split if TL prefers the strict "t/2050 first" sequence.

Co-Authored-By: Azure (Orca) <main.operations-devops-azure@ai-triad-research.orca.local>
Co-Authored-By: Technical Lead (Orca) <main.engineering-tech-lead@ai-triad-research.orca.local>
Co-Authored-By: DevOps (Orca) <main.operations-devops@ai-triad-research.orca.local>

### DIFF
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -166,6 +166,38 @@ jobs:
           tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
 
+      - name: Capture current active revisions (traffic-pin + rollback targets)
+        id: current_revision
+        shell: pwsh
+        run: |
+          # t/2049 + t/2047 landmine: main.bicep defines BOTH the prod and staging
+          # apps, so EVERY apply reconciles BOTH. Capture each app's CURRENT ACTIVE
+          # revision BEFORE the apply and pass both as ingress traffic pins, so the
+          # apply holds traffic on the known-good revision (never auto-promoting a
+          # new one) — for both apps, regardless of which environment we deploy.
+          # The prod capture doubles as the rollback target.
+          $ErrorActionPreference = 'Stop'   # abort on az failure — never pass a
+                                            # silent empty pin when capture itself
+                                            # failed (t/2049#6). Empty is passed ONLY
+                                            # when there is verifiably no active
+                                            # revision (genuine first/only deploy).
+          Import-Module ./scripts/AITriad/AITriad.psm1 -Force
+          function Get-ActivePin {
+            param([string]$App, [string]$Rg)
+            $Active = @(Get-ContainerAppRevision -Mode Active -AppName $App -ResourceGroup $Rg)
+            if ($Active.Count -eq 0) { return '' }
+            # Steady state = exactly one revision at >0% traffic. If a prior deploy
+            # left a split, pin the highest-weight revision.
+            $Top = @($Active | Sort-Object -Property TrafficWeight -Descending)[0]
+            return $Top.Name
+          }
+          $ProdRev    = Get-ActivePin -App '${{ env.CONTAINER_APP_NAME }}'         -Rg '${{ env.RESOURCE_GROUP }}'
+          $StagingRev = Get-ActivePin -App '${{ env.CONTAINER_APP_NAME }}-staging' -Rg '${{ env.RESOURCE_GROUP }}'
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "revision=$ProdRev"
+          Add-Content -Path $env:GITHUB_OUTPUT -Value "staging_revision=$StagingRev"
+          Write-Host "Prod pin / rollback target: '$ProdRev'"
+          Write-Host "Staging pin target:         '$StagingRev'"
+
       - name: Preview Bicep changes (what-if)
         run: |
           echo "Previewing infrastructure changes..."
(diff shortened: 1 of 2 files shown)


jpsnover:
## DevOps co-review: SIGNED OFF (comment, can't approve — same GH account)

**`deploy-azure.yml` — Capture-before-apply:** New step runs BEFORE what-if + arm-deploy. Pin captured while known-good revision is at 100%, passed to Bicep → apply holds traffic there, new revision starts at 0%. Correct ordering.

**`$ErrorActionPreference = 'Stop'`:** Set explicitly in the step. `az` failure throws; empty pin only when `@($Active).Count -eq 0` (verifiably no active revision). t/2049#6 satisfied.

**`Get-ContainerAppRevision -Mode Active`:** Cmdlet confirmed in `scripts/AITriad/Public/Get-ContainerAppRevision.ps1` with `TrafficWeight` property. Sort+first handles split-traffic edge case correctly.

**`maxInactiveRevisions: 10`:** Present on both apps. `$Protected` guard in cleanup is belt-and-suspenders correct.

**Bicep `traffic` ternary:** `empty()` covers both `''` and null. Clean.

**Red-then-safe proof (t/2049#8):** Three-step table (first → unpinned auto-promote → pinned new rev stays 0%) is the correct proof shape. Prod + staging verified untouched.

**deploy.ps1 companion (PR #336):** Open and ready — mirrors the capture + pin logic exactly. t/2049 stays open until it lands per TL instruction. Merge #334 first, then #336.

✅ DevOps sign-off given. Merge on CodeQL green.

PR #575, opened 2026-08-07
feat(shared): add SpeakerIdentity + resolveSpeaker single lookup (t/2242)

jpsnover:
Part 1 of t/2242 (child of t/2234, design critique §2 + §6). **Rosetta Stone to review + land** per the split agreed in e/67 — this adds one new file plus tests and touches no existing shared code. No consumers change here; adoption lands per scope afterward, as Rosetta asked (land-order: component on main before any adopter imports it).

## What

Three surfaces answered "who is speaking" three different ways for the same four-camp taxonomy, with 3+ copies of the label/color lookup, plus a fourth path in diagnostics coloring regex-matched names through its own lowercase map with an `#888` fallback.

`resolveSpeaker(nameOrId)` is the single lookup. It accepts a canonical id, a display label, or a persona alias, in any case.

## The alias table is new, by necessity

Per TL (e/67#4) and confirmed by Rosetta against the soul docs (e/67#5): `Prometheus`/`Sentinel`/`Cassandra` existed **only** inside `DebateExchangeRich.tsx:9`'s segmentation regex — there is no canonical alias map anywhere. So this table becomes the source of truth, with a comment saying future aliases go here rather than into a consumer's regex.

**Behavior change worth calling out:** an unrecognized speaker now resolves to a titlecased label with **no color**, where `debaterColor` returned `#888`. Unstyled reads as "unknown"; a grey pill reads as a real camp. Adoption should expect that difference.

## Everything else composes what exists

- labels + colors → `POVER_INFO` (soul docs) — the same source `speakerLabel`/`speakerColor` already used
- camp + glyph → `povToCamp` + `CampGlyph`
- types → `SpeakerId` via the renderer's `@lib/debate/types` re-export

**One deviation to flag:** TL pointed at the camp type at `lib/debate/types/session.ts:585`, but that is an inline union inside `InterpretationRevisionProposal`, not an exported type. The exported equivalent is `SpeakerId` (`types/phase.ts:6`) with `PovKey`/`POV_KEYS` (`types/synthesis.ts:224`). I used `SpeakerId` and did not redefine anything — flagging in case a different type was meant.

## Variants

`inline` (transcript glyph + colored label), `avatar` (chat circular chip), `badge` (diagnostics solid pill). Each matches what its surface renders today so adoption is a swap, not a restyle. `showLabel={false}` gives glyph-only; a camp-less speaker renders label-only, which is what `INodeRow` needs.

## Verification

- 24/24 new tests: every POV, all three aliases, case/whitespace insensitivity, all four non-POV participants, unknown-speaker degradation, all three variants, glyph-only, badge fill, alias end-to-end, tooltip default + override
- `tsc -p tsconfig.main.json` clean (AC) · eslint 0 errors 0 warnings · depcruise 0 violations · `vite build` clean

### DIFF
--- a/taxonomy-editor/src/renderer/components/shared/SpeakerIdentity.css
+++ b/taxonomy-editor/src/renderer/components/shared/SpeakerIdentity.css
@@ -66,9 +66,17 @@
 .speaker-identity-badge .speaker-identity-label {
   font-size: var(--text-2xs);
   font-weight: 700;
-  color: #fff;
+  color: var(--text-primary);
 }
 
-.speaker-identity-badge .speaker-identity-glyph {
+/*
+ * White reverses out ONLY over a real accent fill. An unknown speaker gets no
+ * fill (resolveSpeaker returns no color), so an unconditional `#fff` here would
+ * print white on `--bg-tertiary` — invisible in the light/harvard/bkc themes.
+ * TL asked us to verify exactly this when endorsing the #888 → no-color change
+ * (e/67#7).
+ */
+.speaker-identity-badge-filled .speaker-identity-label,
+.speaker-identity-badge-filled .speaker-identity-glyph {
   color: #fff;
 }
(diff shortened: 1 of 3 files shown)


jpsnover:
Closing — superseded by #577. #574 (the earlier single-commit duplicate) was merged to main by mistake, which put this branch into CONFLICT so it can't land as-is. #577 restores THIS PR's reviewed content (badge-filled legibility fix + 25 tests) verbatim on top of current main. Content preserved, review stands (e/67#12).

PR #596, opened 2026-08-07
feat(scripts): prototype static token-contrast check (t/2264)

jpsnover:
Prototype half of t/2264. **Not proposed as a gate yet** — DevOps owns wiring, and there is a baseline problem to settle first (below).

## Why this exists

Four contrast defects surfaced in t/2234; three reached main and two shipped live under a default-on flag. All the same shape: **text reversed out over a tokenized fill that is light in some themes.** None was catchable by the suite — vitest runs under jsdom, which applies no stylesheet, so computed color is never evaluated. Every one was found by hand-computing WCAG ratios against `styles.css`.

## Validation — re-derives the real corpus

Run against each defect's **pre-fix** CSS, it independently reproduces the hand-computed figures:

| Defect | Detected | Matches manual |
|---|---|---|
| dark active-step numeral | 2.54:1 | ✅ |
| bkc completed checkmark | 1.36:1 | ✅ |
| white-on-white unfilled badge | 1.13:1 light, 1.31:1 harvard | ✅ |
| filled badge, dark + bkc | 2.39 / 2.60 / 2.99, 3.15 / 3.16 / 3.18 | ✅ (all six) |

Each fixed version produces **no** finding. Seven self-tests (`node --test scripts/check-contrast.test.mjs`) pin both detection and the no-false-positive cases against fixtures.

## Four design decisions that matter

- **Undefined tokens are their own error class.** Treating an unresolvable `var()` as "no background" would reproduce the exact blind spot that hid the white-on-white badge.
- **Ancestor background lookup**, not same-declaration-block only. Both badge defects put the fill on the parent — block-only matching misses half the corpus.
- **Alpha compositing.** Comparing `rgba(C, .12)` against `color: C` raw produces impossible 1:1 ratios; that alone accounted for ~70 false positives.
- **`contrast-fill:` annotation** for fills applied at runtime from data. The filled badge's fill is an inline style — invisible to any CSS-only analysis. This is the only way that defect becomes checkable, and it is opt-in per selector.

## The baseline problem (for DevOps)

Current main yields **278 findings + 36 unresolvable fills**, essentially all pre-existing and outside t/2234. So it cannot go straight to `fail`. Options, in my order of preference:

1. **Ratchet**, like `check-renderer-tsc.sh` (t/2252) — freeze today's count as a ceiling that may only go down.
2. Scope to changed files in the PR diff.
3. Warn-only until the backlog is burned down.

I'd take (1): it matches an existing pattern the team already understands, and it makes every new violation a hard failure immediately.

## Scope note

This lives in `taxonomy-editor/scripts/`, outside my `debate-workspace/` scope — flagged in t/2264#1 and unobjected. Happy to relocate if DevOps or Rosetta prefer it elsewhere; nothing here depends on its location.

Also worth knowing: `scripts/check-camp-contrast.mjs` hardcodes a snapshot of the theme palette rather than reading `styles.css`. It currently agrees with the real tokens, but nothing keeps it honest — and it only checks camp-on-background (which passes), never the reversed-out direction (which failed). It gave coverage that didn't exist. This check subsumes it.

### DIFF
--- a/taxonomy-editor/scripts/__fixtures__/contrast/cases.css
+++ b/taxonomy-editor/scripts/__fixtures__/contrast/cases.css
@@ -1,0 +1,61 @@
+/* Fixture cases for the contrast checker's self-test (t/2264).
+   Each block reproduces one shape from the real t/2234 defect corpus. */
+
+/* PASS — white on a dark fill (same block). */
+.fx-ok-same-block {
+  background: var(--fill-dark);
+  color: #fff;
+}
+
+/* FAIL — white on a light fill (same block). Shape of the phase-indicator glyphs. */
+.fx-fail-same-block {
+  background: var(--fill-light);
+  color: #fff;
+}
+
+/* FAIL — fill on the ancestor, text on a descendant. Shape of the badge defects;
+   a checker matching only same-block pairs misses this entirely. */
+.fx-ancestor {
+  background: var(--fill-light);
+}
+.fx-ancestor .fx-child {
+  color: #fff;
+}
+
+/* PASS — theme-scoped override inverts the glyph where the fill is light.
+   The override must cover EVERY theme whose fill is light: `:root` seeds all
+   four, so leaving bkc/harvard out reproduces the original defect there. That
+   is precisely the mistake the shipped fixes had to avoid. */
+.fx-scoped {
+  background: var(--fill-light);
+  color: #fff;
+}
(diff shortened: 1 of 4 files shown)


jpsnover:
Superseded by PR #597, which includes these prototype commits plus the ci-gate integration (check-contrast.sh, quality-gates.json ceiling 314, ci.yml contrast-check job).

PR #908, opened 2026-08-12
perf(ci): skip post-merge push CI for docs-only changes to main

jpsnover:
## Problem

Every merge to `main` triggers a post-merge push CI run. For **push** events, `github.event_name != 'pull_request'` is always true, so all gated jobs run unconditionally — the full electron matrix + Pester + container smoke (~7 min avg) re-runs even after a 1-line docs merge that already passed those gates on the PR.

This week: at least 8 docs-only PRs each burned a redundant ~7-min CI run on the post-merge push. ~56 minutes of wasted compute per week for zero quality benefit.

## Fix

Add `paths-ignore` to the `on: push:` trigger so pushes where **every** changed file is a markdown/docs path skip the workflow entirely.

## Soundness

- `paths-ignore` only fires when **all** changed files match — a doc commit that sneaks in any `.ts`/`.ps1`/`.json` file falls through to the full suite
- `ci.yml` itself is not listed, so workflow edits always re-run CI
- `deploy/azure/runbooks/**` listed because runbook `.md` files never affect compiled artifacts
- **PR CI behavior is unchanged** — the `changes` path-filter already correctly skips code jobs for docs-only PRs; this closes the equivalent gap on the post-merge push side

## Gate integrity proof needed (TL)

Gate-touching change per AGENTS.md. Requires TL gate-proof with both arms:
1. A docs-only push to main skips the workflow (no spurious CI run)
2. A push containing any code file alongside docs still runs the full suite

## Identified in

Weekly workflow review 2026-08-12.

Co-Authored-By: DevOps (Orca) <main.operations-devops@ai-triad-research.orca.local>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,17 @@ name: CI
 on:
   push:
     branches: [main]
+    # Skip the post-merge re-run for docs-only pushes. All code PRs already
+    # ran CI before merging (branch protection). A push where every changed
+    # file matches this list cannot affect any compiled artifact, so re-running
+    # the electron matrix / Pester / container smoke is pure redundancy.
+    # SOUNDNESS: paths-ignore skips the workflow only when ALL changed files
+    # match — a docs commit that sneaks in a .ts or .ps1 file won't match
+    # and will run the full suite. ci.yml itself is never ignored (not listed).
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'deploy/azure/runbooks/**'
   pull_request:
     branches: [main]
   workflow_dispatch:
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -39,6 +39,9 @@ env:
 jobs:
   ci-gate:
     runs-on: ubuntu-latest
+    permissions:
+      actions: write   # required for gh workflow run (dispatch CI on docs-only tips)
+      contents: read
     steps:
       - name: Verify container-relevant CI passed for this commit
         env:
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -63,8 +63,28 @@ jobs:
             --jq '.workflow_runs[0].id // empty' 2>/dev/null || echo "")
 
           if [ -z "$RUN_ID" ]; then
-            echo "::error::No CI run found for commit ${TARGET_SHA}. Push to main to trigger CI first."
-            exit 1
+            # No CI run found — likely a docs-only tip that paths-ignore skipped.
+            # Dispatch CI manually and wait for it rather than failing with an
+            # unhelpful "push to main" instruction that no longer applies.
+            echo "No CI run found for ${TARGET_SHA} (docs-only tip skipped by paths-ignore). Dispatching CI..."
+            gh workflow run ci.yml --ref main
+            # Allow ~20s for the dispatched run to register in GitHub's API
+            sleep 20
+            for i in $(seq 1 18); do
+              RUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/ci.yml/runs?head_sha=${TARGET_SHA}&per_page=1" \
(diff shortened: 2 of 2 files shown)


jpsnover (commented):
## TL gate-proof review (Main) — mechanism APPROVED, one required change before merge, two-arm proof plan below

**Soundness of the skip semantics — confirmed.** `paths-ignore` skips only when ALL changed files match; `ci.yml` is not listed so workflow edits always run; the `pull_request` trigger is untouched, so required PR contexts (ci-gate, CodeQL) cannot starve — no stuck-PR risk. I also checked breadth: prompt templates are `.prompt` (not `.md`), and nothing in `docs/**` or runbooks is a build/test input. Downstream `workflow_run` consumers (`nudge-container-build`, `deploy-staging`) simply do not fire on a docs-only merge, which is the desired behavior.

**Required change (same PR, DevOps scope): fix the container.yml remediation message.** `container.yml` ci-gate (lines 62-68) looks up the push-triggered ci.yml run by tagged SHA and hard-fails with *"Push to main to trigger CI first."* After this PR, a tag on a docs-only tip fails that lookup and the printed remediation is impossible — a docs push no longer triggers CI. Fail-closed is correct; the wrong instruction during a (typically time-pressured) release is the trap. Update the message to: `gh workflow run ci.yml --ref main` (the dispatched run records head_sha = main tip, satisfying this lookup; ci-gate already accepts test-container "skipped", and dispatch is exercised today by pr-triage). Add a line to the release runbook if one covers this path.

**Two-arm gate proof (post-land — a push-to-main trigger cannot fire pre-merge):**
1. **Run arm — free, immediate:** the merge of #908 itself touches `.github/workflows/ci.yml`, which is not ignored → its merge push MUST produce a ci.yml run. Record the merge SHA + run ID.
2. **Skip arm:** next docs-only merge (or a deliberate one-line docs PR) → `gh api "repos/:owner/:repo/actions/workflows/ci.yml/runs?head_sha=<mergeSHA>"` MUST return empty. Record the SHA + empty result.

Record both pieces of evidence on the ticket per gate-verification policy. With the container.yml message fix pushed, you are cleared to self-merge on green and run the proof.

jpsnover (commented):
## TL re-review of 3e546e81 — one blocker, then cleared

**Gate semantics of the auto-dispatch: approved.** Self-heal inside the gate is fine here because both failure paths still exit 1 and success requires the dispatched run to actually pass (`gh run watch --exit-status`) before falling through to the existing test-container job check — no false-green path. workflow_dispatch is also one of the two event types exempt from GitHub's GITHUB_TOKEN recursion guard, so the dispatched run will genuinely start.

**Blocker — the dispatch will 403 as written.** The workflow-level `permissions` block (container.yml lines 27-32) grants `contents/packages/security-events/id-token/attestations` but **no `actions` scope**. The existing run-lookup reads work only because this repo is public; `gh workflow run` is a write (`POST .../dispatches`) and will fail with 403 under `${{ github.token }}`. Your new remediation path is dead code — a docs-only tip would fail at the dispatch line with a raw 403, not even reaching your crafted messages. Fix: add job-level permissions on `ci-gate`:

```yaml
  ci-gate:
    permissions:
      actions: write   # dispatch + watch ci.yml when paths-ignore skipped the push run
      contents: read
```

Job-level is preferred over widening the workflow-level block — `build-and-push` does not need `actions: write`. (Note job-level permissions REPLACE the workflow block for that job, hence the explicit `contents: read`.)

**Minor, non-blocking:** the dispatch targets `--ref main`, so if `TARGET_SHA` is not main's current tip (tag on an older commit), the dispatched run's head_sha won't match and you'll hit the 3-min timeout with an instruction that also won't help. Worth one clause in the timeout message: "if the target SHA is no longer main's tip, re-run CI via workflow_dispatch is not possible for that SHA — re-tag or build from tip."

Push the permissions fix and you're cleared to self-merge on green; both-arms proof plan unchanged (record evidence on the ticket). Please also verify the dispatch path once live — the first docs-only tag exercise counts as the gate-failure-arm proof for the new remediation.

PR #1020, opened 2026-08-14
gate(build): fail build:server on unemitted dynamic-import-only lib module (t/2626)

jpsnover:
**Build-artifact-completeness gate — t/2626 class closure.** Prevents the news-report failure mode: a dynamic `import()` of a lib/ module doesn't add it to the tsc program, so a module reached ONLY via `import()` silently never emits to dist → `Cannot find module` 500 in the container on first use.

`scripts/check-server-lib-emits.mjs` runs after tsc in `build:server` (so it executes on the real dist in test-container):
- Scans `src/server` for runtime `import('<spec>')` of lib modules; **skips** type-query `import('...').T` forms.
- Maps each specifier — **relative** (`../../lib/X.js`) AND **`@lib`-aliased** — to `dist/server/lib/X.js`.
- **Fails** the build naming the offending import + the missing dist file + the fix (add a static `import type * as X`).
- Scan target + exemptions **co-located** in the script — no external config (TL condition 3).

**Both arms verified locally** (TL conditions p/333#130):
- Clean → `8 dynamic lib import(s) all emitted. OK` (exit 0).
- Fire (hide `dist/server/lib/debate/newsReport.js`) → names `debates.ts:341` + missing file + the fix (exit 1).

**Draft — pending TL Gate Verification** on the real dist. Touches `package.json` `build:server` (Rosetta scope) → flagged for scope review.

Relates t/2626, t/2600.

### DIFF
--- a/taxonomy-editor/package.json
+++ b/taxonomy-editor/package.json
@@ -31,7 +31,7 @@
     "dist:mac": "npm run build && electron-builder --mac",
     "dist:win": "npm run build && electron-builder --win",
     "dist:linux": "npm run build && electron-builder --linux",
-    "build:server": "node -e \"const fs=require('fs');fs.rmSync('dist/server',{recursive:true,force:true});fs.rmSync('.tsbuildinfo/server.tsbuildinfo',{force:true})\" && tsc -p tsconfig.server.json",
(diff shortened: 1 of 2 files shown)


jpsnover:
Rosetta Stone cleared the `package.json` scope review (p/74#114): the `&& node scripts/check-server-lib-emits.mjs` is correctly placed (post-tsc, real dist), `&&`-chained so a miss fails the build, cross-platform (node fs/path), and runs only on build:server/build:container — not the electron build/dev paths. **Remaining gate: TL Gate Verification** (both-arms integrity + exemption-list correctness). Held draft until then.

PR #1039, opened 2026-08-14
fix(nav): Op-Eds reveal honors both build flags; drift-proof navCtx (t/2641)

jpsnover:
**Hardening + the REAL remaining web-reveal fix.** A live smoke showed the web Op-Eds reveal was still broken after #1033. Two drift sites:

1. **navCtx hardcoded flag subset** (Toolbar + HamburgerMenu) — a gate referencing an omitted flag is silently never satisfied (hid Op-Eds in t/2599 + t/2633). Fix: thread the WHOLE flag record (`useFeatureFlagStore(s => s.flags)`) into navCtx. Stable store ref, no t/2142 loop risk.

2. **The real remaining web blocker:** the Toolbar PRIMARY Op-Eds button (line 340) was hardcoded on `env-electron-opeds` ONLY and never used navCtx. #1033 fixed the navCtx path — but in the Toolbar that only feeds the secondary/overflow menu, not the primary button. `App.tsx` always renders `<Toolbar/>` (HamburgerMenu is mobile-only), so **desktop-width web still hid Op-Eds**. Fix: gate the button on `visibleItems.some(i => i.id === 'opeds')` — the same navConfig gate.

Guard test: asserts the Op-Eds button hides with no flag, and shows when ONLY `env-web-opeds` is on (the case line 340 missed) + when only `env-electron-opeds` is on. Fails against pre-fix code.

⚠️ **Deploy note:** this — not #1033 alone — makes the web reveal work on desktop-width web. **The t/2614 smoke needs a redeploy that includes this commit.**

Verify: Toolbar.test.tsx 9/9 · renderer tsc clean · eslint 0 errors. Self-cert (UI-only + guard test); flagging the scope/impact to TL as it's the failure-class prevention area.

### DIFF
--- a/taxonomy-editor/src/renderer/components/shared/HamburgerMenu.tsx
+++ b/taxonomy-editor/src/renderer/components/shared/HamburgerMenu.tsx
@@ -10,7 +10,7 @@ import { SettingsDialog } from '../settings/SettingsDialog';
 import { FeedbackPopover } from './FeedbackPopover';
 import { useAuthStatus } from '../../hooks/useAuthStatus';
 import { useBreakpoint } from '../../hooks/useBreakpoint';
-import { useFlag } from '../../hooks/useFeatureFlags';
+import { useFlag, useFeatureFlagStore } from '../../hooks/useFeatureFlags';
 import { getGlobalRecorder } from '@lib/flight-recorder/index';
 import {
   NAV_ITEMS, getVisibleNavItems, getSecondaryByGroup,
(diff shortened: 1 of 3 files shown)


jpsnover:
Superseded. #1035 (already on main) did the navCtx whole-flag-record threading via a useNavVisibilityContext hook — better than this branch's inline version, and it conflicts. Re-cut the remaining delta (the primary Op-Eds button honoring the gate + guard test) as a minimal PR off current main.

PR #1043, opened 2026-08-14
feat(server): getStateRoot() — isolate class-A writes from the shared data root (t/2643)

jpsnover:
**Prod-integrity fix (t/2643).** Staging + prod mount the SAME Azure Files share, so any class-A write from staging (feature flags, runtime config, keys, calibration) mutated prod's data. Splits the writable **STATE root** from the read/grounding **DATA root**. Per TL GV (t/2643#9) + SO conditions: the RO mount DevOps flips on staging is the PRIMARY boundary; this split makes staging function within it.

**config.ts:** `getStateRoot()` = `AI_TRIAD_STATE_ROOT ?? getDataRoot()` (prod/local unchanged); `resolveStatePath()`; `isStagingIdentity()` (drift-proof `CONTAINER_APP_NAME`); `assertStateRootIsolation()` boot fail-fast (staging && stateRoot===dataRoot → refuse start); keyStore wiring → `getStateRoot` (12 sites, keyStore.ts unchanged).

**Rerouted to `getStateRoot()` (raw-fs class-A, read+write together):** featureFlags (flags+audit), runtimeConfig, providerBinding, admin telemetry, flight-recorder dirs (server.ts + diagnostics), calibration (debates.ts:53 call site — lib helper root-agnostic, no DebateTool change; diagnostics calibration reads).

**Stays on `getDataRoot()` (shared RO):** grounding/taxonomy + read-only provisioned config (quotas, proxy-tiers, authorized-users). Backend-mediated writes (writeDump, calibrationStore, user content) → GitHub-API/Blob in prod, not the mount → separate class-B question.

**Folded (TL p/333#158):** `/api/data/pull` refuses on staging (its `reset --hard`/`clean -fd` runs on the shared worktree). Prod requireAdmin + .gitignore audit = **t/2645** (separate PR).

**Tests:** `stateRoot.test.ts` (11) — default/override, isStagingIdentity, fail-fast both ways, + the unit-level **arm-B property** (a flag write lands in the state root, data root untouched). featureFlags.test.ts (16) unchanged = backward compat. tsc + eslint clean; server suite green (75 local fails = Node-7.x undici-skew, CI green).

**DRAFT — routed to Main (TL) for Gate Verification** (both runtime arms on the RW mount: staging debate-smoke + config + key write → prod `/mnt/shared` byte-unchanged; staging write to RO mount fails visibly) before merge + the RO flip. DevOps half: RO-mount shared share, RW staging-state mount, Bicep-declared `AI_TRIAD_STATE_ROOT` + `TAXONOMY_CACHE_DIR` on both apps.

Relates t/2643, t/2644, t/2645, t/2612.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/stateRoot.test.ts
+++ b/taxonomy-editor/src/server/__tests__/stateRoot.test.ts
@@ -1,0 +1,97 @@
+// @vitest-environment node
+//
+// t/2643 — state-root isolation. Class-A WRITES (feature flags, runtime config, keys,
+// calibration, telemetry, flight-recorder) route through getStateRoot(); grounding READS stay
+// on getDataRoot(). On staging, AI_TRIAD_STATE_ROOT points at an isolated RW mount so a class-A
+// write cannot mutate prod's shared data root. These tests prove the split + the fail-fast guard
+// + the unit-level arm-B property (a flag write lands in the state root, NOT the data root).
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { getDataRoot, getStateRoot, resolveStatePath, isStagingIdentity, assertStateRootIsolation } from '../config.js';
+import { setFlag, getFlagMetadata, _resetFlagCache } from '../featureFlags.js';
+
+const SAVED: Record<string, string | undefined> = {};
+const KEYS = ['AI_TRIAD_DATA_ROOT', 'AI_TRIAD_STATE_ROOT', 'CONTAINER_APP_NAME', 'AI_TRIAD_ENV', 'ADMIN_USERS'];
+let dataRoot: string;
+let stateRoot: string;
+
+beforeEach(() => {
+  for (const k of KEYS) SAVED[k] = process.env[k];
+  dataRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'data-'));
+  stateRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'state-'));
+  for (const k of KEYS) delete process.env[k];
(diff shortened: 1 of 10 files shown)


jpsnover:
**GV confirmations (TL t/2643#14), on record:**

**(1) saveProposal (fileIO.ts:937) exclusion is correct.** `saveProposal` writes via the module-level `backend` (`getBackend()`). Default is `FilesystemBackend` (fileIO.ts:45), but in prod `github-api` mode server.ts:236-242 runs `fileIO.setBackend(new GitHubAPIBackend(...))` at boot — so the active backend is **GitHubAPIBackend → writes to the GitHub API, not `/mnt/shared`**. Excluding it from getStateRoot is correct (class-B/backend surface, not the raw-fs mount hazard).

**(2) analytics (server.ts:1462, left on getDataRoot) fs-fallback can't fire on staging.** server.ts:1459-1463 passes `{accountUrl}` **iff** `AZURE_STORAGE_ACCOUNT_URL` is set → Azure Append Blobs (no fs write); the local-NDJSON path fires only when it's UNSET. Staging has `AZURE_STORAGE_ACCOUNT_URL` set (same as prod — it's what makes user content Blob-backed), so analytics → Blob, never the fs fallback. Defense-in-depth: even if it were unset, the RO mount → EACCES → the init `.catch` swallows it (analytics disabled, **no prod mutation**). *DevOps to confirm the env is set on staging.* Optional one-line hardening if you'd rather not depend on the env: repoint `initAnalytics(getStateRoot())` so a fired fallback stays isolated — say the word and I'll add it, else it stays as approved.

Merge gate acknowledged: alongside DevOps's Bicep (RO-last), runtime both-arms with FULL triggers (flag+config+key+debate+proposal-save+analytics → `/mnt/shared` byte-unchanged). Un-drafting when DevOps's half is ready + t/2644/t/2645 PRs linked.

PR #1082, opened 2026-08-15
docs(review): add Class 6 - silent degradation under single-context validation (t/2669)

jpsnover:
## What

Adds **Class 6 — Silent degradation under single-context validation** to the failure-class taxonomy (`docs/CodeReview/failure-classes.md`). Completes **t/2669 AC4**.

## Why

The 2026-08-15 dual-build / staging-isolation sweep surfaced a recurring class the existing five didn't name. It's the amplifier class: a change validated in **one** context (local / one build target / one env / one identity) fails in an un-validated context, and a graceful-degradation path — empty result, swallowed error, success-return — **hides** the failure in prod. It passes every green check while visibly broken.

## The prevention (three moves, always together)

1. Verify in the **deployed** context (hosted web / github-api path, real deployed data, anon identity) — not just local + authenticated.
2. Assert the **positive outcome** — data presence (count > 0 / expected shape), not "renders / responds without error."
3. Make degradation **observable** — graceful-empty (ADR-001) and swallow-and-continue must emit a log/metric signal.

## Evidence cited (this session)

- Dual-build read parity — 26/26-green smoke while Entities/Orgs empty on web (t/2648/t/2661)
- Swallow-and-continue — analytics append (t/2664), route-layer log-map bug (#1065)
- Deploy/env false-greens — durability warm-pod (t/2642), staging data-root at empty cache mount (t/2670)
- Concurrency swallow — feature-flag clobber returning success (t/2644)

Docs-only. The companion process rule (AC2) already landed in the tech-lead `AGENTS.md` **Cross-Profile Impact**.

### DIFF
--- a/docs/CodeReview/failure-classes.md
+++ b/docs/CodeReview/failure-classes.md
@@ -1,10 +1,10 @@
 # Failure-Class Taxonomy — Prospective Review Checklist
 
-**Last updated:** 2026-08-09
+**Last updated:** 2026-08-15
 **Owner:** Diagnostics (source analysis) / Tech Lead (review checklist)
-**Source:** e/84 systemic quality analysis of the 2026-08-09 Azure production bug sweep.
+**Source:** e/84 systemic quality analysis of the 2026-08-09 Azure production bug sweep; Class 6 added from the 2026-08-15 dual-build / staging-isolation sweep (t/2669).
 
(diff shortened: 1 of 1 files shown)


jpsnover:
Superseded — the identical Class 6 (silent degradation under single-context validation) landed via #1083 (merge commit 06909b78) at 18:40, all checks green. This branch now conflicts against main because the same failure-classes.md section is already present. Closing as a duplicate — parallel-work collision on t/2669 AC4 (both instances implemented it independently; #1083 merged first). No content lost: Class 6 is on main. Apologies for the double-open.

PR #1095, opened 2026-08-15
fix(security): route /api/fetch-url through the hardened SSRF fetcher (t/720 -> t/2482)

jpsnover:
## Problem

The repo has **two** SSRF-guarded URL fetchers, and the internet-facing Azure route used the weaker one.

- **Strong (t/2482):** `lib/url-fetch/fetchUrlForPrompt.ts` — IP-pinned connect, size cap, timeout, content-type gate, per-hop redirect re-validation. Backs the debate source path.
- **Weak (t/720):** `taxonomy-editor/src/server/storage/urlFetch.ts` — reached by `POST /api/fetch-url` via `fileIO`. **This is what runs on Azure.**

`urlFetch.ts` pre-checked DNS and then called `fetch(url)`, which re-resolves the hostname at connect time — a DNS-rebinding TOCTOU window. Its own comment called this "residual… minimized", accurate when written, but t/2482 has since demonstrated the real fix (connect to the validated IP, preserving SNI + `Host`). The route returns the response body to the caller, so a bypass is a full read primitive.

**Exposure:** authenticated-only (`/api/fetch-url` is in neither `PUBLIC_EXACT_PATHS` nor `ANON_SAFE_POST_PATHS`), but prod runs `AUTH_OPTIONAL='1'` and `enforceOptionalAuthGate` never consults the allowlist — so the bar is "holds a Google/GitHub/Microsoft account", not "is on the allowlist".

**Blast radius, scoped honestly:** this is *not* a managed-identity credential path. Container Apps doesn't serve IMDS at `169.254.169.254`; tokens come from a `127.0.0.1` `IDENTITY_ENDPOINT` over http with a required `X-IDENTITY-HEADER`. The HTTPS-only check and the 127/8 block each independently prevent it. What a rebind reaches is private HTTPS endpoints inside the ACA environment — internal probing/scanning. The environment declares no VNet, so egress is unrestricted and app-layer validation is the only control.

## Change

`urlFetch.ts` no longer implements transport. It keeps only what is specific to this route and delegates the fetch:

1. A **stricter** pre-flight than the shared fetcher applies — HTTPS-only, no credentials, no local/internal suffixes. (The shared fetcher permits `http:` for the debate path; this route must not, so the pre-flight stays.)
2. markitdown HTML→Markdown conversion, preserved via a new `rawBody` option on the shared fetcher.

`rawBody` skips the tag-strip/sanitize pass and returns the body verbatim so markitdown converts the original markup. Transport-level guarantees are unaffected — there is a test asserting the SSRF block still fires under `rawBody`.

The route inherits, all previously absent: **10s timeout**, **1.5 MB cap with mid-stream abort**, **content-type gate**, **per-hop redirect re-validation** (was 1 hop).

### Also in this PR

- **`lib/url-fetch/**/*.test.ts` was absent from the vitest `include` list** — the hardened fetcher's suite has never gated CI since t/2482 landed. Added, with a `node` environment pragma (it binds real loopback servers; the suite defaults to jsdom). Worth noting independently of this change: that module was already in production on the debate path, untested in CI.
- `Host` header now carries a non-default port (RFC 9110 §7.2) — it was being dropped.
- `http-error` results carry the status, so callers still surface `HTTP 404` rather than a generic string.
- Failed fetches record to the flight recorder; the pre-consolidation code recorded DNS failures and that hook would otherwise have been lost.

## Behavior change worth reviewing

The content-type gate now rejects non-text URLs with a clear error. Previously a PDF link was `resp.text()`-ed into mojibake, written to `input.html`, and handed to markitdown as HTML — garbage either way, so **not a functional regression**, but users pasting arXiv PDF links now get an explicit refusal instead of nonsense. Making that path actually work (markitdown does handle PDFs) is a reasonable follow-up.

## Tests

New coverage:
- **IP-pinning proof** — a `.invalid` hostname (RFC 2606, never resolves) reachable *only* if the transport connects to the address from the pre-flight lookup. Also asserts the Host header + port fix. This is the rebind regression test; it lives at the lib level because once the IP is pinned a rebind is structurally unobservable at the server level — there is no second resolution to intercept.
- `rawBody` passthrough, and SSRF-still-blocks-under-`rawBody`.
- `http-error` carries status.
- Server-level: bracketed IPv6 literal rejected pre-DNS; and a guard that domain names starting `fc`/`fd`/`ff` are **not** misread as IPv6 literals — `isPrivateIp` classifies IPv6 by string prefix, so an ungated literal check would have blocked `fdsa.com`, `ffmpeg.org`.

Results (rebased onto `origin/main` @ 78f5c4a8, deps reinstalled):
- `t720Security.test.ts` + `lib/url-fetch/` — **40 passed, 0 failed**
- `npm run build:server` (incl. `check-server-lib-emits`) — clean, 8 dynamic lib imports emitted
- `depcruise` — no violations across 1596 modules (the new server→`lib/url-fetch` edge is legal)
- `eslint` on changed files — clean

**Full server suite locally: 1091 passed, 76 failed.** All 76 are in `githubApi.test.ts` and `t2020Security.test.ts`, from the undici major-skew guard firing on local Node (bundled undici 7.x vs the required Node 22 / undici 6.x). The identical 76 fail on unmodified `main`, so they are pre-existing and environmental — this branch adds exactly +2 passing tests. CI pins node:22.23.2 and should be green; that is the one claim I could not verify locally.

## Notes

No ticket key: the Orca MCP server rejected `create_ticket` in the authoring session (`Cannot determine caller agent identity` — `ORCA_AGENT_ID` unset), so the commit references `t/720 → t/2482`. Happy to amend the message if a ticket is filed.

Routing: the `taxonomy-editor` sub-roles (ServerAPI / Server Storage) are not registered in the Orca workspace — `resolve_owner` on `urlFetch.ts` falls back to the root project role. Per AGENTS.md, security-surface work is TL-reviewed by default.

### DIFF
--- a/lib/url-fetch/fetchUrlForPrompt.test.ts
+++ b/lib/url-fetch/fetchUrlForPrompt.test.ts
@@ -1,7 +1,12 @@
 // Copyright (c) 2026 Jeffrey Snover. All rights reserved.
 // Licensed under the MIT License. See LICENSE file in the project root.
 
+// @vitest-environment node
+// Binds real loopback HTTP servers and drives node:http/node:dns — the suite's
+// default jsdom environment cannot host either.
+
 import * as http from 'node:http';
+import { promises as dnsPromises } from 'node:dns';
 import { AddressInfo } from 'node:net';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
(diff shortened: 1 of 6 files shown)


jpsnover:
## TL security review — APPROVED (merge on green). Security surface, so routed to TL per AGENTS.md.

Reviewed the full diff as an SSRF/OWASP change. This is the right consolidation, not just a wrapper.

**The actual vuln it closes:** the internet-facing `POST /api/fetch-url` used a *second, weaker* transport — DNS pre-check then `fetch(url)`, which re-resolves the hostname at connect time = a DNS-rebinding TOCTOU. This PR deletes that path and delegates to the single hardened `fetchUrlForPrompt` (t/2482), which connects to the **pre-validated IP** (closes the rebind window) and adds a request timeout, response size cap w/ mid-stream abort, content-type gate, and **per-hop** redirect re-validation — none of which the removed path had.

**Verified:**
- Call site `fetchUrlForPrompt(url, { rawBody: true })` uses the DEFAULT `checkAddress = isPrivateIp` → resolved-IP private/internal block is enforced; `rawBody` only skips post-fetch tag-strip (markitdown converts the markup itself). Lib test confirms SSRF block still fires with `rawBody`.
- Stricter route pre-flight preserved: HTTPS-only, no credentials, reject `localhost`/`.local`/`.internal`/`.corp`, bracketed-IPv6 literal rejected pre-DNS.
- **Real false-positive fixed:** the old `isPrivateIP(hostname)` + prefix-based IPv6 classification blocked any host starting `fc`/`fd`/`ff` (fdsa.com, ffmpeg.org). New code gates the literal check on `isIpLiteral()`; the resolved-IP check remains authoritative (defense-in-depth). Regression test covers it.
- **Gate blind spot closed:** `vite.config.ts` adds `lib/url-fetch/**/*.test.ts` to CI — the hardened fetcher's suite wasn't gating CI since t/2482 (Class 5). Good catch.
- Observability: ssrf-blocked recorded as `warn`, others `info` — blocks are detectable, not silent.

**Both arms present:** internal/rebind URLs blocked (ssrf-blocked), external allowed (ALLOW_ALL happy-path). WHATWG URL normalizes octal/hex/decimal IPv4 to dotted-quad, and the resolved-IP backstop catches anything the literal check misses.

**Disposition:** APPROVED. Merge once CI is green (currently running). No further TL review needed.

PR #1160, opened 2026-08-17
fix(oped): block news-hook fabrication + wire source-claim grounding

jpsnover:
## Summary

- **t/2721** — Empty `newsHook` fallback now explicitly forbids inventing dated events (pending votes, rulings, named regulations). CL-approved guard text directs the model to open on enduring stakes or SOURCE MATERIAL instead.
- **t/2722** — Source-claim grounding pipeline: `OpEdGroundingRef.document_claims?`, `SOURCE_KEY_CLAIMS` numbered list in user prompt, reflection prompt/schema extended with `SOURCE_CLAIMS` block and per-element `document_claims` (all CL-approved verbatim). `GenerateOpEdRequest.sourceBrief` → `sourceMaterial` rename. `assembleReflectionPrompt` gains `sourceClaims` default param — existing call site unchanged until TL wires Step-0 claim-reflection pass.

## Changed files

- `lib/oped/types.ts` — `OpEdGroundingRef.document_claims?: string[]`
- `lib/oped/promptLoader.ts` — fabrication-blocking fallback, `SOURCE_KEY_CLAIMS` interpolation, `assembleReflectionPrompt` signature
- `lib/oped/generate.ts` — `sourceMaterial` rename, `REFLECTION_SCHEMA` + `ReflectionResponse` extended
- `lib/oped/prompts/op-ed-generation-system.prompt` — §1 lede/news-hook guidance (CL verbatim)
- `lib/oped/prompts/op-ed-generation-user.prompt` — `SOURCE_KEY_CLAIMS` block added
- `lib/oped/prompts/op-ed-grounding-reflection.prompt` — full replacement with `SOURCE_CLAIMS` + `document_claims` (CL verbatim)
- `lib/oped/promptAssembly.test.ts` — 4 new tests; 32/32 pass

## Test plan

- [ ] CI green on head
- [ ] CL reviews wiring: numbered-list rendering for `SOURCE_KEY_CLAIMS`/`SOURCE_CLAIMS`, clean empty-source degradation
- [ ] TL confirms `sourceMaterial` rename is compatible with Step-0 orchestration branch

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/oped/generate.ts
+++ b/lib/oped/generate.ts
@@ -251,7 +251,8 @@ async function runVoiceGeneration(
   if (allGroundingRefs.length > 0 && body) {
     try {
       const groundingList = buildGroundingList(groundingNodes, sitNodes);
-      const reflPrompt = assembleReflectionPrompt(deps.promptsDir, body, groundingList);
+      // sourceClaims stays '(none)' for P1 — key_claims require the comprehension pass in Step-0 (TL)
+      const reflPrompt = assembleReflectionPrompt(deps.promptsDir, body, groundingList, '(none)');
       const reflMaxTokens = Math.max(4000, allGroundingRefs.length * 150 + 3000);
       const reflRaw = await deps.adapter.generateText(reflPrompt, request.params.model, {
         maxTokens: reflMaxTokens,
@@ -260,10 +261,13 @@ async function runVoiceGeneration(
         signal: request.signal,
       });
       const reflParsed = JSON.parse(stripCodeFences(reflRaw)) as ReflectionResponse;
-      const usageMap = new Map(reflParsed.grounding_usage.map(u => [u.id, u.reflection]));
+      const usageMap = new Map(reflParsed.grounding_usage.map(u => [u.id, u]));
       for (const ref of allGroundingRefs) {
-        const refl = usageMap.get(ref.node_id);
-        if (refl) ref.how_reflected = refl;
+        const usage = usageMap.get(ref.node_id);
+        if (usage) {
+          ref.how_reflected = usage.reflection;
+          if (usage.document_claims?.length) ref.document_claims = usage.document_claims;
+        }
       }
     } catch {
       // Reflection failure is non-fatal — how_reflected stays '(not reported)'
--- a/lib/oped/generate.ts
+++ b/lib/oped/generate.ts
@@ -21,8 +21,8 @@ export interface GenerateOpEdRequest {
   topic: string;
   params: OpEdParams;
   povs: PovKey[];
-  /** P2 slot: pre-fetched source brief text (FromUrl). Omit for P1 FromTopic. */
-  sourceBrief?: string;
+  /** Raw source markdown (pre-fetched from URL). Omit for topic-only requests. */
+  sourceMaterial?: string;
   signal?: AbortSignal;
 }
 
@@ -158,6 +158,7 @@ const REFLECTION_SCHEMA = {
         properties: {
           id: { type: 'string' },
           reflection: { type: 'string' },
+          document_claims: { type: 'array', items: { type: 'string' } },
         },
         required: ['id', 'reflection'],
       },
@@ -178,7 +179,7 @@ interface EssayResponse {
 }
 
 interface ReflectionResponse {
-  grounding_usage: { id: string; reflection: string }[];
+  grounding_usage: { id: string; reflection: string; document_claims?: string[] }[];
 }
 
 async function runVoiceGeneration(
@@ -201,7 +202,7 @@ async function runVoiceGeneration(
     voiceBlock: buildVoiceBlock(soul),
     groundingNodes: formatGroundingNodes(groundingNodes),
     situations: formatSituationNodes(sitNodes),
-    sourceMaterial: request.sourceBrief ?? '(no external source supplied — argue from the topic and general knowledge)',
+    sourceMaterial: request.sourceMaterial ?? '(no external source supplied — argue from the topic and general knowledge)',
     outletGuidance: band.guidance,
     targetWords,
   });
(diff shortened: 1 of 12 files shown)


jpsnover:
## CL wiring review — PR #1160 (t/2721 + t/2722)

**Verdict: `needs-changes`** (draft — do not un-draft until the t/2722 wiring below is closed). The **t/2721 fabrication fix is clean and can land as-is**; the **t/2722 source-claim grounding is scaffolded but inert at runtime** — the prompt/schema/types plumbing is in place, but nothing flows through it because three population sites are missing. 32/32 green is real but the suite only exercises the rendering functions in isolation (`ctx.sourceBrief` set directly in the test); no test drives the `generate.ts` path that would reveal the gaps.

Reviewed the two items you flagged plus the end-to-end path.

### ✅ What's right
- **News-hook fallback (t/2721)** — `promptLoader.ts` fallback + system-prompt §1 both drop "invent a plausible" and replace with enduring-stakes framing + explicit "do NOT invent a dated event." Good wording, correctly conditioned on SOURCE MATERIAL presence. `promptAssembly.test.ts` asserts it.
- **Numbered-list rendering logic** — both `SOURCE_KEY_CLAIMS` (`c, i) => \`  ${i+1}. ${c}\``) and the reflection `SOURCE_CLAIMS` slot are formatted correctly *when given data*.
- **Empty-source degradation** — clean. Empty → `SOURCE_KEY_CLAIMS` = `(none extracted)`, `SOURCE_CLAIMS` = `(none)`, `sourceMaterial` = argue-from-topic fallback. No blank interpolation, no fabrication invite. This item passes.

### ⛔ Issues (blocking for t/2722; none block t/2721)

**Issue 1 (severity: blocking, category: wiring) — `SOURCE_CLAIMS` is never populated.**
`generate.ts:253` calls `assembleReflectionPrompt(deps.promptsDir, body, groundingList)` with no 4th arg, so `sourceClaims` takes its `'(none)'` default on **every** run. The reflection model is therefore always told "Empty if this op-ed had no source document," so it will always return empty `document_claims`. The numbered-list rendering for `SOURCE_CLAIMS` you asked me to check is correct but unreachable.
*Fix:* thread the source brief's `key_claims` (numbered) into the call. In-scope for this PR (pure `lib/oped`).

**Issue 2 (severity: blocking, category: wiring) — reflection `document_claims` is dropped, so `OpEdGroundingRef.document_claims` never populates.**
The reflection loop builds `usageMap` from `u.reflection` only; `u.document_claims` is read nowhere and `ref.document_claims` is never assigned. So the new `types.ts` field + schema + prompt additions are dead — the BDI-element→document-claim subchildren the user expected (t/2722 AC #2) never render, even with Issue 1 fixed.
*Fix:* map `u.document_claims` onto each ref alongside `how_reflected`. In-scope for this PR (pure `lib/oped`).

**Issue 3 (severity: blocking, category: wiring — likely split across roles) — structured `sourceBrief` is never built or passed; comprehension pass not invoked.**
The context handed to `loadAndAssemblePrompt` carries only `sourceMaterial` (raw markdown) — there is no `sourceBrief` key, and `GenerateOpEdRequest` no longer has a structured-brief field at all. So `SOURCE_KEY_CLAIMS` and every `SOURCE_AUTHOR/THESIS/STANCE/RECOMMENDATIONS` field render empty in production, and `op-ed-source-brief.prompt` is still never called from `generateOpEdSet` (the original t/2722 root cause). This is t/2722 AC #1 and is the larger piece — it spans `lib/oped` (invoke the brief pass, populate `PromptContext.sourceBrief`) and the handlers (t/2722 is root-assigned). **Please confirm the intended split:** is Issue 3 this PR or a follow-up? Issues 1–2 should be in this PR regardless, since they make the pieces this PR *does* add functional.

### Suggestions (non-blocking)
- **S1 (t/2721 AC #2/#4):** `FABRICATED_LEDE_GUARD` currently lives only inside the test file and is asserted against literal strings — it never scans a generated lede, so it can't catch a real fabrication in prod. Fine as a stipulated first step, but consider exporting it and applying it to generated output (or file a follow-up for the N≥10 empty-hook regression in AC #4). I (CL) owe the `metric-provenance-register` entry for this guard (stipulated lexicon) — I'll take that as a CL follow-up once the pattern is final.
- **S2:** if you pass the numbered list into the reflection `SOURCE_CLAIMS` slot, the prompt says copy claims "verbatim from the numbered list" — the model may then copy the `1. ` prefix into `document_claims[]`. Either pass claims **unnumbered** to the reflection slot, or tell it to copy the claim text without the leading number, so `document_claims` stays clean for later matching against `key_claims`.

### Recommendation
- Split intent: land the **t/2721** fabrication fix (it's complete and correct). For **t/2722**, fix Issues 1–2 in this PR (small, pure `lib/oped`) and confirm whether Issue 3 (comprehension pass + structured brief) is this PR or a tracked follow-up. Keep the PR **draft** until that's settled — the `document_claims` schema/prompt/types are inert without at least Issues 1–2.
- I'll file CL follow-ups for the provenance-register entry (S1) and can draft the numbered→reflection threading text if useful.

jpsnover:
**CL follow-up review — verdict now `approve`.** Re-reviewed `876582af`.

- **Gap 2 — fixed correctly.** `usageMap` maps the full usage object; `ref.document_claims = usage.document_claims` only when non-empty, matching the optional-field "absent when no source brief" semantics. Reflection loop is sound.
- **Gap 1 — accepted as deferred.** Explicit `'(none)'` + comment is the right call: `key_claims` don't exist until the Step-0 comprehension pass, so there's nothing to thread yet. The pipe is wired and dormant.
- **Gap 3 — confirmed follow-up** under t/2722 AC#1 (TL Step-0 scope).

Net: the source-claim grounding is complete-and-dormant — it activates automatically once Step-0 populates `key_claims` and flips `sourceClaims` off `'(none)'`. t/2721 fabrication fix + Gap 2 are the live wins here.

**One suggestion (non-blocking):** no test asserts `document_claims` propagation through the reflection loop (existing grounding tests mock `assembleReflectionPrompt`). Worth a small test feeding a reflection response with `document_claims` and asserting it lands on the ref — cheap regression insurance for the now-dormant path. Good to un-draft once CI's green on this head.

PR #1165, opened 2026-08-17
feat(oped): wire FABRICATED_LEDE_GUARD to production + guard-regression tests

jpsnover:
## Summary

- `lib/oped/opedGuards.ts` — exports `FABRICATED_LEDE_GUARD` as single source of truth (canonical form, stipulated in metric-provenance-register §5 / t/2726)
- `lib/oped/types.ts` — adds `fabricated_lede?: true` to `OpEdMember`
- `lib/oped/generate.ts` — scans first 500 chars of generated body against guard when `newsHook` was empty; sets `fabricated_lede: true` + flight-recorder warning on match (no body mutation — flag+warn, CL-approved)
- `lib/oped/promptAssembly.test.ts` — imports guard from `opedGuards.ts`, removes inline copy
- `lib/oped/opedGuards.test.ts` — 12 positive cases + 12 timeless-framing negative cases (guard-regression mock, N≥10 per AC#3)
- `lib/oped/serialGeneration.test.ts` — 3 guard integration cases (clean/empty-hook, fabricated/empty-hook, fabricated/non-empty-hook)

40/40 tests pass.

## Test plan

- [ ] CI green on head
- [ ] CL reviews + signs off; graduates register row §8→§5

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/oped/generate.ts
+++ b/lib/oped/generate.ts
@@ -13,6 +13,8 @@ import { computeEmbedding } from '../embeddings/onnxEmbedding.js';
 import type { OpEdMember, OpEdParams, OpEdSet, OpEdGroundingRef } from './types.js';
 import { resolveOutletBand } from './outletBands.js';
 import { loadAndAssemblePrompt, assembleReflectionPrompt, type SourceBrief } from './promptLoader.js';
+import { FABRICATED_LEDE_GUARD } from './opedGuards.js';
+import { getGlobalRecorder } from '../flight-recorder/index.js';
 
 // ── Public request / deps types ───────────────────────────────────────────────
 
@@ -274,6 +276,17 @@ async function runVoiceGeneration(
     }
   }
 
+  // Guard scan: when newsHook was empty, check the lede (first 500 chars) for
+  // fabricated dated-event markers. Flag without mutating the body (t/2730).
+  const emptyHook = !request.params.newsHook?.trim();
+  const fabricatedLede = emptyHook && FABRICATED_LEDE_GUARD.test(body.slice(0, 500));
+  if (fabricatedLede) {
+    getGlobalRecorder()?.record({
+      type: 'system.warning', component: 'opedGenerate', level: 'warn',
+      message: `FABRICATED_LEDE_GUARD matched for pov=${pov} set=${request.set_id} — empty-hook lede may contain invented dated event`,
+    });
+  }
+
   return {
     pov,
     status: 'complete',
@@ -283,6 +296,7 @@ async function runVoiceGeneration(
     pitch: parsed.pitch_email || undefined,
     wordCount: actualWordCount,
     grounding: allGroundingRefs,
+    ...(fabricatedLede && { fabricated_lede: true as const }),
   };
 }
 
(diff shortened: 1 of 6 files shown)


jpsnover:
## CL review — HOLD (CI red, not the flaky kind)

Implementation faithfully matches the approved design (flag+warn, no body mutation; guard exported to `opedGuards.ts` as single source; `promptAssembly.test.ts` de-duped; negative arm correctly labeled "guard-regression (mock)"; good integration arm incl. the non-empty-hook suppression case). **But I can't approve or graduate the register while CI is red.**

**Two real failures on `09c15c37`:** `renderer-tsc` and `test-server-prod-config`. Both are deterministic (a typecheck + a prod-config build), so this is not the install flake we hit on #1160.

**Root-cause hypothesis (strong, but confirm with the exact `tsc` error — the run was still in progress so I couldn't pull the failing line):**
- `lib/oped` had **zero** recorder usage before this PR (`grep recorder lib/oped/*.ts` on main = 0 hits). This PR adds the first one: `import { getGlobalRecorder } from '../flight-recorder/index.js'` in `generate.ts`.
- The renderer + server builds compile `generate.ts` by import-following from their entry points, so that new static edge pulls `flight-recorder/index.js` (a broad barrel) into build graphs that didn't include it before. That's almost certainly what `renderer-tsc`/`test-server-prod-config` are choking on. (Note: the renderer tsconfig `include` lists flight-recorder but not `lib/oped` — the coupling arrives via import resolution, not the include list.)

**Recommended fix (cleaner than chasing the tsc error): thread the recorder through `OpEdGeneratorDeps`** (DI, alongside `adapter`/`promptsDir`/`repoRoot`) instead of importing the global singleton. Benefits: (1) avoids the new cross-package static import edge that's breaking the downstream builds; (2) matches the existing dependency-injection pattern in this module; (3) removes the need to `vi.mock('../flight-recorder/index.js')` in `serialGeneration.test.ts` — that mock is itself a smell that the global import doesn't belong here. The caller (opedHandlers / web route) passes the recorder it already has.

Please paste the exact `tsc`/build error so we confirm the mechanism, apply the fix, and get to green. Once green I'll re-verify and graduate the register §8→§5.

jpsnover:
## CL — still HOLD, now on a merge conflict (not CI)

The DI fix (`776787ec`) is **correct** — global import removed, `recorder?` threaded through `OpEdGeneratorDeps`, call site `deps.recorder?.record(...)`, `vi.mock` gone. That was exactly the ask.

But the PR is now **`mergeable: CONFLICTING` / mergeState DIRTY**, so GitHub isn't running CI on `776787ec` ("no checks reported on the branch" — checks are gated behind conflict resolution, not missing).

**Conflicts vs current `main`** (merge-base `56141dcb`; `main` advanced via a landed op-ed PR touching these):
- `lib/oped/generate.ts` — changed in both (their edit overlaps the guard-scan block you added)
- `lib/oped/serialGeneration.test.ts` — changed in both

**Action (your branch):** merge/rebase `origin/main` into `feat/oped-fabrication-guard`, resolve those two files (keep both the guard scan + the incoming changes), push. That clears DIRTY and triggers CI on the new head. Once green I'll re-verify the four gates, merge, and graduate the register §8→§5.

PR #1262, opened 2026-08-19
feat(t2808): schema version-bump CI gate for lib/brief/schemas

jpsnover:
## t/2808 — schema version-bump gate

Diffs the three `lib/brief/schemas/*.write.json` write schemas against the PR base and flags any **contract-surface change** (properties/required/enum/pattern/type) that lands without bumping the schema's `$id` version — the version-bump discipline the forward-compatible IR (T1, Second Opinion e/109#2) cannot self-enforce (on-disk manifests would misreport their contract).

### Mechanics (self-cert per /trivial-change, as the ticket permits)
- `lib/brief/schemas/check-version-bump.mjs` — pure logic (`schemaVersion`, `contractSignature`, `detectViolation`) + a git-driven CLI. Config **co-located** with the schemas (AC).
- Description/title/`$id`/`$schema` are stripped before comparing, so doc-only edits and pure version bumps never trip it.
- `--mode=warn` (default: GH `::warning`, exit 0) / `--mode=block` (`::error`, exit 1). Promote to block after a green warn cycle (Gate Promotion).
- 13 unit tests (added-prop-no-bump → violation; added-prop-with-bump → clean; description-only → clean; version-only → clean; required/enum change → violation; new/deleted file → ignored).

### Both-arms Gate Verification evidence (local) — **TL runs the official GV**
- **Clean** (no schema change): `clean (3 schemas)`, exit 0.
- **Red / block** (property added, same version): `::error`, exit 1, "Failing the build".
- **Red / warn**: `::warning`, exit 0, "Warn-only for now".

### Remaining (coordination)
- **DevOps**: wire a `deckspec-version-bump` CI step (warn mode first) into `ci.yml` — `node lib/brief/schemas/check-version-bump.mjs` with the PR base fetched (`BASE_REF`/`GITHUB_BASE_REF`). ci.yml is DevOps-owned; draft step in the ticket.
- **TL**: both-arms Gate Verification, then promote warn→block after ≥1 green real-PR cycle.

lib eslint clean; 13 tests pass.

### DIFF
--- a/lib/brief/schemas/check-version-bump.mjs
+++ b/lib/brief/schemas/check-version-bump.mjs
@@ -1,0 +1,158 @@
+#!/usr/bin/env node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// Brief Export schema version-bump gate (t/2808).
+//
+// Under the forward-compatible IR (T1, Second Opinion e/109#2), the version-bump
+// discipline is a SOCIAL rule the schema can't self-enforce: a `properties` change
+// to any lib/brief/schemas/*.write.json with no version bump silently erodes the
+// provenance guarantee (on-disk manifests would misreport which contract they were
+// written under). This gate diffs the three write schemas against the PR base and
+// fails when a contract change lands without bumping the schema's `$id` version.
+//
+// The version marker is the trailing `/<major>.<minor>` of each schema's `$id`
+// (e.g. .../deck_spec/1.0). Additive-optional → bump minor; breaking → bump major.
+//
+// Config is co-located with the schemas (this file lives beside them). Pure logic
+// is exported for unit tests; the CLI runner drives it against `git`.
+//
+// Usage: node lib/brief/schemas/check-version-bump.mjs [--mode=warn|block]
+//   --mode=warn  (default) prints violations as GitHub warnings, exits 0.
+//   --mode=block prints errors, exits 1 on any violation (promote here after a
+//                green warn cycle — Gate Promotion, t/2808 AC).
+// Base ref: $BASE_REF, else origin/$GITHUB_BASE_REF, else origin/main.
+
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL Gate Verification — Script PASS ✅ (one pre-promotion condition)

Ran the full both-arms GV against the real write-schemas in a worktree at this branch head:

| Arm | Result |
|---|---|
| Clean (no contract change vs `origin/main`) | `clean (3 schema(s) checked)`, exit 0 — zero noise ✅ |
| Violation @ `--mode=warn` (property added, no `$id` bump) | `::warning::` annotation, exit 0 (non-blocking) ✅ |
| Violation @ `--mode=block` | `::error::` annotation, exit 1 ✅ |

Discrimination is correct: description/title-only edits and version-bumped contract changes are unit-tested as non-violations; `contractSignature` strips `$id`/`description`/`title` so the version marker itself isn't part of the surface. Config is co-located with the schemas (Gate Co-Location ✅). Warn-first is the right promotion posture (Gate Promotion ✅).

**Script is cleared to merge** (warn phase). One condition, tracked against the `--mode=block` promotion, **not** this merge:

### Pre-promotion condition: `gitShow()` swallows all git errors → false-pass
`gitShow()` returns `null` on *any* `git show` failure, and `detectViolation()` treats `base == null` as "new schema → OK". That conflates two very different cases:
- **genuinely new schema file** (legit — no prior contract), vs
- **`git show` errored** (unfetched base ref, transient git failure) → should fail **loud**, not silently pass.

In warn phase with the base ref fetched this is low-risk. But before promoting to `--mode=block`, harden it: on a `git show` error where the ref *resolves* but the path lookup fails (vs. a clean "path absent at a valid ref"), the gate must error rather than report clean — otherwise a git hiccup = silent non-enforcement, which is the exact false-pass class this gate exists to prevent. (This is also why #1264 must set `fetch-depth: 0` — see that PR.)

PR #1264, opened 2026-08-19
ci: schema version-bump warn gate (t/2808)

jpsnover:
## Summary
- Adds `schema-version-bump` job to `ci.yml` (warn mode, `continue-on-error: true`)
- Runs `node lib/brief/schemas/check-version-bump.mjs --mode=warn` on PRs touching `lib/**`
- Exits 0 always — GitHub warning annotations only, no CI gate breakage
- Script file lands in PR #1262 (t/2808); without it the step fails gracefully (file not found, continue-on-error catches it)

## Gate Promotion path (NOT in this PR)
When promoted to `--mode=block`: remove `continue-on-error`, add `schema-version-bump` to `ci-gate needs`, open as a separate draft PR + TL sign-off per Gate Promotion Discipline.

## Test plan (TL GV)
- [ ] Arm 1 (violation): PR that changes `*.write.json` contract without bumping `$id` → step prints `::warning::` annotation, exits 0, job passes
- [ ] Arm 2 (clean): PR with no schema changes (or bumped version) → step prints "clean", exits 0

Blocked on TL GV before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -523,6 +523,31 @@ jobs:
         working-directory: lib
         run: pnpm exec tsc --noEmit -p debate/tsconfig.json
 
+  # ── Schema version-bump gate — warn mode (t/2808) ───────────────────────
+  # Diffs lib/brief/schemas/*.write.json against the PR base; warns when the
+  # contract surface changes without a $id version bump. Warn-only (exit 0
+  # always) until a green warn cycle confirms signal quality — then promote to
+  # --mode=block + add to ci-gate (Gate Promotion per AGENTS.md, requires TL
+  # sign-off per DevOps AGENTS.md Gate Promotion Discipline).
+  schema-version-bump:
+    needs: [changes]
+    if: >-
+      always() && !cancelled() &&
+      (github.event_name != 'pull_request' || needs.changes.outputs.lib == 'true')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v6
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v6
+        with:
+          node-version: '22'
+
+      - name: Schema version-bump check — warn mode (t/2808)
+        continue-on-error: true
+        run: node lib/brief/schemas/check-version-bump.mjs --mode=warn
+
   # ── Commit hygiene: flags anonymous / undersized commit subjects (t/1319) ──
   # Annotation-only by design — never flips (t/1319). Visibility, not friction.
   commit-hygiene:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -539,6 +539,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v6
+        with:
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL Gate Verification — CHANGES REQUIRED 🔴

The gate **script** (#1262) passes both-arms GV locally. But this **CI wiring cannot fire** as written — I verified it empirically.

### Defect: shallow checkout → the gate is a silent no-op

The step's `actions/checkout` has **no `fetch-depth: 0`**. The default is a shallow depth-1 fetch, so `origin/<base>` is **absent on the runner**. The gate diffs via `git show origin/<base>:<path>`; when that ref is missing, `gitShow()` catches the error and returns `null`, and `detectViolation()` treats every schema as "new" → **no violation is ever reported**, even in `--mode=block`.

**Proof** (real violation present — a property added to `deck_spec.write.json` with no `$id` bump):

```
# base ref UNREACHABLE (simulates the shallow CI checkout)
$ BASE_REF=origin/this-ref-not-fetched node check-version-bump.mjs --mode=block
fatal: invalid object name 'origin/this-ref-not-fetched'.   (x3, swallowed)
schema version-bump gate: clean (3 schema(s) checked).
exit=0        ← SILENT FALSE-PASS

# base ref reachable (control)
$ BASE_REF=origin/main node check-version-bump.mjs --mode=block
::error ... changed without bumping the schema $id version ...
exit=1        ← fires correctly
```

Consequence beyond "no signal": a green warn cycle here would be **clean because the gate is blind, not because PRs are compliant** — that fraudulent green would then be used to justify promotion to `--mode=block`, where it would *still* never fire.

### Required before merge
- Add `with: { fetch-depth: 0 }` to the `actions/checkout` step (fetches all branches/tags so `origin/main` is present).
- Re-run and confirm the gate can **see the base** — ideally push a throwaway commit that changes a `*.write.json` property without a version bump and confirm the `::warning::` annotation actually appears in the job log (real-env failure-arm, per Gate Promotion). A warn cycle that never produced a single annotation is not evidence the gate works — it's evidence it ran.

Script GV verdict is on #1262. Ping me when `fetch-depth: 0` is in + you've seen it annotate — fast re-clear.

jpsnover:
## TL Gate Verification — CLEARED for warn-phase merge ✅

`fetch-depth: 0` verified on the checkout step (commit 87fc742b). That closes the silent-no-op defect — the gate can now reach `origin/<base>` and, as proven in the local both-arms GV, it fires correctly whenever the base is reachable. Warn-mode + `continue-on-error: true` is the right non-blocking posture. **Good to merge.**

### Two hard preconditions before promoting to `--mode=block` (not now):
1. **Observed real-env annotation.** During the warn cycle, land one deliberate schema-contract change with no `$id` bump and confirm the `::warning::` actually appears in the job log. A warn cycle that only ever printed "clean" is evidence the gate *ran*, not that it *works* — do not treat clean-only as the green cycle.
2. **`gitShow` hardening** (see #1262): a `git show` error must fail loud, not fall through to "new schema → clean". With `fetch-depth: 0` the base is present, but a transient git failure must not silently disable enforcement in blocking mode.

Both are TL-signed-off gates on promotion per the DevOps Gate Promotion Discipline. Ship the warn wiring now.

jpsnover:
Hold: waiting for PR #1262 (script file) to merge before un-drafting. TL GV cleared at p/331#434. Block-promotion preconditions: (1) observe real ::warning:: annotation in CI during warn cycle; (2) gitShow error-hardening in #1262. Neither required before this PR merges.

PR #1267, opened 2026-08-19
ci: lib/brief NodeNext typecheck step in lib-lint (t/2832)

jpsnover:
## Summary
- Adds `pnpm run typecheck:brief-nodenext` step to the existing `lib-lint` job
- Script: `tsc -p brief/tsconfig.nodenext.json` (from PR #1265, t/2832)
- Path-filtered to `lib/**` via existing `lib` filter — runs on any lib change
- Step is blocking (no `continue-on-error`) — deterministic typecheck, no flake surface
- Prevents the #1259 break shape: NodeNext module resolution errors that pass standard tsc

## Dependencies
- Requires PR #1265 to merge first (adds `typecheck:brief-nodenext` script to `lib/package.json`)

## Test plan (TL GV)
- [ ] Arm 1 (violation): PR with a lib/brief NodeNext incompatibility → step fails, lib-lint red
- [ ] Arm 2 (clean): normal PR touching lib/** → step passes, no noise

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -523,6 +523,10 @@ jobs:
         working-directory: lib
         run: pnpm exec tsc --noEmit -p debate/tsconfig.json
 
+      - name: TypeScript — lib/brief NodeNext module check (t/2832)
+        working-directory: lib
+        run: pnpm run typecheck:brief-nodenext
+
   # ── Schema version-bump gate — warn mode (t/2808) ───────────────────────
   # Diffs lib/brief/schemas/*.write.json against the PR base; warns when the
   # contract surface changes without a $id version bump. Warn-only (exit 0


jpsnover:
## TL Gate Verification — CLEARED for blocking ✅ (two merge-order conditions)

Both arms of the underlying check are verified on #1265 (clean→0/exit0; ajv-default regression→TS2351 not-constructable/exit2). Wiring here is sound:

- **Deps present before the step.** `lib-lint` runs `pnpm install --frozen-lockfile` (line ~516) before this step (~524); the existing `debate/tsconfig.json` typecheck in the same job already proves tsc resolves lib deps there, and my clean-arm run confirmed ajv/pptxgenjs/etc. all resolve for `typecheck:brief-nodenext`. No fetch-depth-style precondition gap (this is a pure tsc — no base-ref/network/deploy dependency, so offline both-arms IS the real-env behavior; the warn-first-real-env promotion concern doesn't bind here).
- **Zero tolerated noise** — clean arm is a true 0.

### Conditions before merge (ordering, not redesign)
1. **Merge after #1265** — this step calls `typecheck:brief-nodenext`, which #1265 defines. Merging first = the blocking step fails on a missing script and blocks every lib PR.
2. **Green on this PR's own CI first** — confirm the step passes in the real `lib-lint` job (the clean-arm-in-real-env cycle) before it goes blocking on `main`.

Good to go blocking once both hold.

jpsnover:
Hold: lib-lint fails because typecheck:brief-nodenext script isn't in main yet (comes from PR #1265). Merge #1265 first, then rebase this branch and re-run CI. TL GV cleared at p/460#1.

PR #1269, opened 2026-08-19
feat(t2816): allowOpen / meta.snapshot support in extractDeckSpec

jpsnover:
## t/2816 — end the allowOpen deferral (lib part)

`extractDeckSpec(session, opts?: { allowOpen?: boolean })`. A non-closed debate still throws **by default** (unchanged); with `allowOpen` it exports a **watermarked in-progress snapshot** — sets `meta.snapshot=true` + `meta.snapshot_note`. T4 already renders the "IN PROGRESS" watermark + maturity note on `meta.snapshot`, so this makes the snapshot path reachable end-to-end on the lib side. Additive + backward-compatible (optional param; snapshot fields already optional in the schema; closed path byte-identical).

### Design decisions (for review)
- **Concluding-entry guard relaxed for snapshots:** an in-progress debate legitimately has no concluding synthesis → synthesis-derived sections are empty rather than a throw (the "content may be incomplete" reality). Closed still requires it.
- **meta.phase mapping:** `DebateSession.phase` is a session lifecycle enum (`setup|clarification|edit-claims|opening|debate|closed|cancelled`), distinct from the deck_spec `concluding|closed|open`. Map `closed→closed`, any non-closed→`open` ('concluding' has no lifecycle equivalent). Previously hard-coded 'closed' (only reachable when closed).
- **`cancelled` is not special-cased** — a cancelled debate + allowOpen would still snapshot. Flagging in case you want to exclude it; easy to add.

### Tests (5 new, 42/42 pass)
default non-closed still throws; snapshot sets flag+note+phase=open; missing-concluding tolerated under allowOpen; closed unchanged (no snapshot flag); closed+allowOpen is a no-op. Clean under nodenext (the new t/2832 gate) + bundler; lint clean.

### Fast-follows (other scopes — flagging, not in this PR)
- **T6 (root Main):** flip `routes/briefExports.ts` `?allowOpen=true` 400 → run the pipeline (extract now accepts it), surface the maturity warning.
- **T8 (PowerShell):** `-AllowOpenDebate` format gate.

→ Main-TL: contract change to extract (you reviewed it for T2) — routing for review.

### DIFF
--- a/lib/brief/extract.test.ts
+++ b/lib/brief/extract.test.ts
@@ -122,6 +122,46 @@ describe('extractDeckSpec phase guard', () => {
   });
 });
 
+// ── allowOpen / meta.snapshot (t/2816) ────────────────────────────────────────
+
+describe('extractDeckSpec allowOpen snapshot', () => {
+  it('non-closed still throws by default (allowOpen omitted)', () => {
+    const session = makeSession({ phase: 'debate' });
+    expect(() => extractDeckSpec(session as never)).toThrow("must be 'closed'");
+  });
+
+  it('non-closed + allowOpen exports a snapshot with meta.snapshot + note + phase open', () => {
+    const session = makeSession({ phase: 'debate' });
+    const spec = extractDeckSpec(session as never, { allowOpen: true });
+    expect(spec.meta.snapshot).toBe(true);
+    expect(spec.meta.snapshot_note).toContain('snapshot');
+    expect(spec.meta.snapshot_note).toContain('debate'); // the real session phase
+    expect(spec.meta.phase).toBe('open');
+  });
+
+  it('allowOpen tolerates a missing concluding entry (in-progress snapshot)', () => {
+    const session = makeSession({ phase: 'opening', transcript: [] });
+    const spec = extractDeckSpec(session as never, { allowOpen: true });
+    expect(spec.meta.snapshot).toBe(true);
+    // synthesis-derived sections are legitimately empty before the debate concludes
+    expect(spec.agreements).toEqual([]);
+    expect(spec.cruxes).toEqual([]);
+  });
+
+  it('closed export sets no snapshot flag (unchanged)', () => {
+    const spec = extractDeckSpec(makeSession() as never);
+    expect(spec.meta.snapshot).toBeUndefined();
+    expect(spec.meta.snapshot_note).toBeUndefined();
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL review — APPROVED ✅ (extract layer clean; two fast-follow flags)

Verified the change against the schema, not just the green suite.

**No schema change / version bump needed — and that's correct.** The T1 write schema *already* declares `meta.phase: enum [concluding, closed, open]`, `meta.snapshot: boolean`, and `meta.snapshot_note` (the forward-compatible IR from the Second Opinion ruling anticipated this). extract is only *populating* already-declared optional fields, so `meta.additionalProperties:false` strict-on-write still passes — that's why 42/42 are green, not a validation gap. The t/2808 version-bump gate correctly does not fire (schema unchanged). Genuinely additive/backward-compat, confirmed.

**Design decisions — all sound:**
1. **Concluding-entry relaxation** — correctly gated: `if (!concludingEntry && !isSnapshot) throw`. The **closed path keeps its strictness** (still errors on a missing concluding entry); only snapshots tolerate it (synthesis → `{}` → empty sections). No regression. ✓
2. **phase mapping** — `closed ? 'closed' : 'open'`, real session phase preserved verbatim in `snapshot_note`. Fine. Minor optional: the enum also allows `'concluding'`; you collapse it to `'open'`. Acceptable as-is (the note carries fidelity) — refine only if a concluding-phase deck should read `'concluding'`.
3. **cancelled not special-cased** — reasonable; a cancelled debate exports as a snapshot with the note. ✓

**Two flags for the T6/T8 fast-follows (NOT blocking this PR — the extract layer correctly emits the data):**
- **T5 verify must be snapshot-aware.** A snapshot with empty synthesis sections must not *hard-fail* the verify gate's coverage/symmetry checks — for `meta.snapshot === true`, those expectations should soft-warn, not fail. Confirm when the T6 flip wires allowOpen through verify, else every snapshot export errors at the gate.
- **T4 watermark must actually consume `meta.snapshot`.** The `snapshot_note` says "T4 renders the IN PROGRESS watermark" — confirm the renderer reads `meta.snapshot` and draws it. If that's not landed, a snapshot exports looking *final* (silent degradation — the exact class our dual-build rule guards). Track it with the flip.

Extract change **approved to merge**. Please make the two flags explicit on the T6/T8 flip tickets so they aren't lost.

PR #1305, opened 2026-08-20
feat(t2852): PDF print + framingMeta toggle for brief export dialog

jpsnover:
## Summary
- Adds classroom-only **framingMeta** checkbox to `BriefExportDialog` (include/exclude framing & meta slide); sent at top-level of POST body
- Adds Electron-only **Save as PDF…** button in done phase — downloads `brief.html` artifact and prints via `printToPDF` offscreen BrowserWindow
- Full IPC chain: `brief:html-to-pdf` handler → preload → `electron.d.ts` → electron-bridge delegation → web-bridge browser-print fallback → `AppAPI`/`BriefExportRequest` types
- CSS: `.bx-pdf-row`, `.bx-pdf-error` (theme-token-safe per t/2567 gate)

## Test plan
- [ ] Web: classroom preset shows framingMeta checkbox; unchecking omits slide (verify server receives `framingMeta: false` at top-level)
- [ ] Web: non-classroom presets show no framingMeta checkbox
- [ ] Electron: done phase shows "Save as PDF…" button only when `htmlDoc` artifact is present
- [ ] Electron: clicking saves file via native save dialog
- [ ] Web: "Save as PDF…" button absent (Electron-only gate)
- [ ] TSC clean: `npx tsc -p tsconfig.json --noEmit` → 0 errors

Closes t/2852

### DIFF
--- a/taxonomy-editor/src/main/ipc/debateHandlers.ts
+++ b/taxonomy-editor/src/main/ipc/debateHandlers.ts
@@ -212,6 +212,32 @@ export function registerDebateHandlers(): void {
     return { cancelled: false, filePath };
   });
 
+  // ── Brief HTML → PDF (t/2852) ────────────────────────────
+  // Render brief.html to PDF via an offscreen BrowserWindow, mirroring debateToPdf.
+  ipcMain.handle('brief:html-to-pdf', async (_event, html: string) => {
+    const { canceled, filePath } = await dialog.showSaveDialog({
+      title: 'Save brief as PDF',
+      defaultPath: 'brief.pdf',
+      filters: [{ name: 'PDF Files', extensions: ['pdf'] }],
+    });
+    if (canceled || !filePath) return { cancelled: true };
+
+    const pdfWindow = new BrowserWindow({
+      show: false,
+      width: 1024,
+      height: 768,
+      webPreferences: { offscreen: true, sandbox: true, contextIsolation: true },
+    });
+    try {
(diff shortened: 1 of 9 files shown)


jpsnover:
**TL fix pushed (`77720f00`) — the "Save as PDF" button was unreachable.**

Root cause (dual-build parity): the button was gated `isElectronMode()`, but the brief-export pipeline is **web-only** in `electron-bridge` (`createBriefExport`/`list`/`downloadBriefArtifact` all reject as web-only). So in Electron the dialog never reaches the `done` phase and `record` never exists — while in web the button was hidden. Net: the feature was reachable in **neither** profile, and the author's own `web-bridge.printBriefToPdf` (browser print) was dead code.

Fix: gate the button on `record.artifacts.includes('brief.html')` only. `api.printBriefToPdf` already dispatches per profile (browser print on web, native `printToPDF` on Electron), so it now works in web (where the pipeline runs) and stays correct on Electron if the pipeline lands there later. Also collapsed a dead `if/else` in `savePdf` into `finally`, dropped the now-unused `isElectronMode` import, and corrected stale "Electron-only" comments.

Regression tests (web mode — previously uncovered): button appears + routes through the bridge when `brief.html` is present; absent when it isn't. `renderer-tsc` 0, `eslint` 0 errors, dialog suite **7/7**.

Note: PDF still requires the pipeline to actually emit `brief.html`. Verify on the hosted web profile against real deployed data before sign-off (dual-build data-read parity).

PR #1327, opened 2026-08-20
fix(brief/render): normalize ppt/presentation.xml to canonical order — verify OOXML gate (t/2871)

jpsnover:
## t/2871 — normalize ppt/presentation.xml to canonical order (T6 release-blocker)

**T6 release-blocker** (TL p/342#116): every real brief export — offline CLI **and** T6 server — hard-fails the verify OOXML gate, because pptxgenjs 3.12.0 emits `<p:sldIdLst>` before `<p:notesMasterIdLst>` on any deck with speaker notes (assemble.ts attaches `speakerNotes` to every slide), violating ECMA-376 `CT_Presentation`. Latent since verify shipped (#1256): no test ever rendered notes through verify.

### Fix — shared render path (covers CLI + server in one shot)
- **`lib/brief/render/normalizeOoxml.ts`** — reorders the `<p:presentation>` id-list/size children into canonical order via pure JSZip/OOXML text surgery (mirrors `potxHonor.ts`). No new dep, no pptxgenjs-surface change, **no image decoding** (t/2866 safe). Returns the bytes unchanged (no re-zip) when already canonical.
- **`render/index.ts`** — applied as the FINAL byte step, after the optional `.potx` merge. Since both CLI (`runBriefPipeline`) and T6 (`briefExportJobs.ts` → same `runBriefPipeline`) go through `render()`, one fix covers both.

### Tests — red-first proven
- `reorderPresentationChildren` unit tests: out-of-order → canonical; already-canonical → unchanged (no re-zip); <2 canonical children → untouched; children + trailing non-canonical elements preserved, none dropped/duplicated.
- **render→verify-WITH-notes integration test** — the exact #1256 gate gap. **Red** on unfixed render (`PptxLintFailure: … <p:sldIdLst> precedes <p:notesMasterIdLst>`); **green** after: the full export is verify-clean (`hardFailures` empty).

### ⚠ Design note (deviation flag)
TL said "fix in shared **pptxRenderer**". I homed the logic in a sibling **`normalizeOoxml.ts`** rather than in `pptxRenderer.ts`, because it is **post-processing (JSZip), not pptxgenjs API** — keeping `pptxRenderer.ts` the *sole* pptxgenjs surface (the same reason `potxHonor.ts` is separate). Wired into the shared `render()` so CLI + server are both covered exactly as intended. Happy to inline into pptxRenderer if you'd prefer.

### Gates
nodenext ✓ · renderer-tsc 0/0 ✓ · vitest 57/57 (normalizeOoxml + render + verify + pipeline; existing suites unaffected) ✓ · `/review-ts` clean. No new runtime dep.

### Follow-up (t/2871 AC)
Once this lands, tighten the t/2868 CLI E2E (`cli.entrypoint.test.ts`) from the anti-silent-no-op contract to the **success-only** assertion (per TL p/342#117).

Routing to **Main (TL)** for review — release-blocker + render-byte-contract change. Not self-merging.

Closes t/2871.

### DIFF
--- a/lib/brief/render/index.ts
+++ b/lib/brief/render/index.ts
@@ -13,6 +13,7 @@ import type { SlideKind } from './slideModel.js';
 import { resolveTheme } from './deckTheme.js';
 import { renderPptx } from './pptxRenderer.js';
 import { renderHtml } from './htmlRenderer.js';
+import { normalizePresentationOrder } from './normalizeOoxml.js';
 
 export interface RenderInput {
   spec: DeckSpec;
@@ -66,6 +67,13 @@ export async function render(input: RenderInput): Promise<RenderResult> {
       allWarnings.push(`template_masters_not_merged: slide-master injection failed — ${msg}`);
     }
   }
+
+  // FINAL byte step: normalize ppt/presentation.xml child order to ECMA-376 canonical
+  // (pptxgenjs emits notesMasterIdLst after sldIdLst on any deck with speaker notes,
+  // which the T5 OOXML lint rejects — t/2871). After the optional .potx merge so it
+  // also corrects any reordering that introduced. No-op (bytes unchanged) if already
+  // canonical.
+  pptxBytes = await normalizePresentationOrder(pptxBytes);
 
   const htmlDoc = renderHtml(slides, theme);
 
--- a/lib/brief/render/normalizeOoxml.test.ts
+++ b/lib/brief/render/normalizeOoxml.test.ts
@@ -1,0 +1,122 @@
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// Tests for the ppt/presentation.xml canonical-order fix (t/2871).
+//   1. Pure unit tests of the reorder logic.
+//   2. Red-first render→verify-WITH-notes integration test — the exact #1256 gate
+//      gap (no test ever rendered a deck with speaker notes through verify), which
+//      is why the pptxgenjs ordering bug stayed latent.
+
+import { describe, it, expect } from 'vitest';
+import { reorderPresentationChildren } from './normalizeOoxml.js';
+import { runBriefPipeline } from '../pipeline.js';
+import type { DebateSession } from '../../debate/types.js';
+import type { AIAdapter } from '../../debate/aiAdapter.js';
+
+// pptxgenjs's buggy shape: <p:sldIdLst> BEFORE <p:notesMasterIdLst>.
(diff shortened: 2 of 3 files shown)


jpsnover:
## TL Review — APPROVED (design/correctness/determinism/security), merge on full green

Reviewed the diff end-to-end. Clean, well-scoped fix. Verdict: **approve**, merge once CI is fully green on 7000fa79.

### Deviation — ACCEPTED, and it's the better call
Homing the logic in a sibling `normalizeOoxml.ts` instead of literally inside `pptxRenderer.ts` is approved — my "fix in pptxRenderer" was shorthand for "in the shared render path," and post-processing OOXML via JSZip is categorically **not** a pptxgenjs-API concern. Keeping `pptxRenderer.ts` the sole pptxgenjs surface (mirroring `potxHonor.ts`) is more aligned with the t/2866 bounded-surface principle than inlining. Wiring it as the FINAL byte step after the `.potx` merge is right — it also normalizes any reorder the merge introduces.

### Determinism — VERIFIED empirically (the load-bearing concern)
`verify.ts:107` hashes `sha256(pptxBytes)` into the manifest and there's no date-pinning in the renderer, so the JSZip re-zip you added on the **always** path (every real deck has notes) had to be proven byte-stable. JSZip's `file()`/`generateAsync()` default entry dates to `new Date()` — the classic ZIP non-determinism trap. Probe result: two `normalizePresentationOrder` calls on identical input **across a >1s wall-clock boundary → byte-identical output** (same sha256). The replaced `presentation.xml` entry is not getting a wall-clock stamp; the round-trip is a deterministic function of input bytes. Good.

### Correctness / security
- Reorder logic verified: canonical check passes (notesMasterIdLst precedes sldIdLst); unit tests cover child-preservation, no-drop/no-dup, already-canonical no-op (`===`, skips re-zip), <2-children untouched. Literal-string `replace` — safe.
- t/2866: JSZip treats entries as opaque bytes — no image decoding, no new pptxgenjs surface reached. Matches potxHonor precedent.
- Test quality: red-first render→verify-**with-notes** integration test closes the exact #1256 gap and asserts `hardFailures === []` broadly. Solid.

### Conditions before merge
1. **Full CI green on 7000fa79** — `test-electron (taxonomy-editor)` (runs the new lib/brief tests) still pending as I write; must pass.
2. `test-electron (summary-viewer)` reported fail while the run was still in progress — zero dependency on lib/brief, almost certainly flake. Re-run; ping me only if it reproduces with an assertion tied to this change.
3. Follow-up (t/2871 AC): tighten the t/2868 CLI E2E to success-only, now unblocked.

Yours to land on green per pre-self-merge verify.

PR #1335, opened 2026-08-20
fix(t2874): Export-TriadDebateBrief — kill exit-0-no-output false-green + always pass --model

jpsnover:
## t/2874 — Export-TriadDebateBrief cmdlet→CLI correctness (t/2868 follow-up)

Two bugs, **both surfaced by a real cmdlet→CLI E2E** that the mocked suite + the false-green had masked:

### 1. Exit-0-no-output false-green
The CLI could exit 0 with **no stdout** (t/2868 Windows `invokedDirectly` misfire) → the wrapper parsed empty stdout into an all-null `TriadDeckExport` and reported success (`"Exported brief:"` no path). Now empty/whitespace stdout — or a null parse — raises `RenderFailure`. Mocked test asserts exit-0 + empty stdout → error, not an empty export. Same guard added to `Test-BriefNarrationStage` (#1333).

### 2. `-SkipNarration` was fully broken end-to-end
It passed `--skip-narration` but **not `--model`**, yet the CLI **requires `--model` on every run** (records it even in deterministic mode) → CLI exited **2** (arg error). Now `--model` is always passed (sentinel `deterministic` when skip + no `-Model`); `--skip-narration` is additive. The old unit test that asserted "`--model` absent under skip" **encoded the bug** — corrected.

### Real E2E (the durable "assert output" coverage t/2874 asked for)
`tests/Export-TriadDebateBrief.E2E.Tests.ps1` runs the **actual `tsx` CLI** against a committed fixture debate and asserts a populated `TriadDeckExport` + a real `brief.pptx` on disk. Skips cleanly when node deps are absent (never a flaky blocking gate).

**Verified live** against the real CLI (shared checkout): `-SkipNarration` with and without `-Model` both now produce a real brief (`traceCov=100`, `brief.pptx` written). Unit suite **21 green**, E2E skips without deps.

Closes the false-green class for the brief cmdlets. Self-cert: scoped bug fix. CLI-side `invokedDirectly` fix is #1325 (Shared Lib).

### DIFF
--- a/scripts/AITriad/Public/Export-TriadDebateBrief.ps1
+++ b/scripts/AITriad/Public/Export-TriadDebateBrief.ps1
@@ -330,7 +330,12 @@ function Export-TriadDebateBrief {
             & $WriteExportError 'RenderFailure' "Output directory is not empty: $OutDir — use -Force to overwrite." $OutDir; return
         }
 
-        $resolvedModel = if ($SkipNarration) { '(none — narration skipped)' } else { $Model }
+        # The CLI REQUIRES --model on every run (it records the model field even in
+        # deterministic mode); -SkipNarration is additive, NOT a replacement for it.
+        # Omitting --model made the CLI exit 2 (arg error), so -SkipNarration was fully
+        # broken end-to-end (t/2874 E2E). Under skip without a model, record the sentinel
+        # 'deterministic' — the CLI accepts it and never calls a model.
+        $resolvedModel = if ($Model) { $Model } elseif ($SkipNarration) { 'deterministic' } else { $Model }
         Write-Verbose "Local mode: debate '$ResolvedPath'"
         Write-Verbose "  Preset=$Preset, model=$resolvedModel$(if ($CheckerModel) { ", checker=$CheckerModel" })$(if ($AllowOpenDebate) { ', allow-open snapshot' }). Output → $OutDir"
         if (-not $PSCmdlet.ShouldProcess($ResolvedPath, "export brief (preset=$Preset, model=$resolvedModel) → $OutDir")) { return }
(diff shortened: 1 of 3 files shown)


jpsnover:
## TL diagnosis + sign-off (f1add33a)

Independently confirmed the root cause — it's a **test-infra bug, not a product bug**:

- The `test-powershell` CI job installs deps with `pnpm install --frozen-lockfile --filter taxonomy-editor` — **only** the taxonomy-editor package, not the root workspace where `lib/brief` lives. So `pptxgenjs`/`jszip` aren't linked for the brief CLI, but pnpm still creates a root `node_modules` dir + `npx` exists → the old `HasDeps` guard passed → the E2E ran → `tsx lib/brief/cli.ts` crashed on a missing import → exit 1 with a Node stack (no `{errorCode}` JSON) → cmdlet reported the generic "brief CLI exited with code 1". "node_modules present" ≠ "pipeline runnable" — exactly what f1add33a's message says.
- **Product verified working:** I ran the identical CLI invocation (`--path <exp-1438 fixture> --preset conference --out … --model deterministic --skip-narration --allow-open`) from a full checkout → **exit 0**, real `brief.pptx` written, only advisory WARN lines. The cmdlet + `-SkipNarration`/`--model` sentinel fixes are correct.
- **Fix validated:** with `AITRIAD_RUN_BRIEF_E2E` unset, the E2E now **skips cleanly** (`Skipped: 1, Failed: 0`, loud reason) — so `test-powershell` goes green instead of false-failing.

Approach (opt-in env gate) is sound and owner-authored (PowerShell). Merge on green.

### One coverage note (non-blocking)
The opt-in gate means this E2E **never runs automatically in CI** (no CI job has both the PS module and lib/brief's runtime deps). The mocked suite + the exit-0-no-output guard remain the CI-enforced coverage, so this is acceptable — but t/2874's original goal of *durable CI output-assertion* is only met locally/on-demand. If we want it enforced in CI, the follow-up is a DevOps ci.yml change: add a job (or extend one) that does a full-root `pnpm install` + runs this E2E with the env var set. Flag me and I'll decompose it as a separate ticket.

PR #1379, opened 2026-08-22
fix(prompts): catalog directional-judge.prompt to unblock main (t/2834)

jpsnover:
**Unblocks broken main** (PM-authorized, p/15#107). `data/promptCatalog.lint.test.ts` (t/2834, forward `.prompt` arm) is failing on `main` — red for **every open PR** — because `scripts/AITriad/Prompts/directional-judge.prompt` (landed via t/2900 / t/2911) was never catalogued or excluded.

## Fix
It''s a first-class PS content prompt — a directional-agreement judge for the polarity gate, invoked by `Invoke-DirectionalJudge.ps1` (t/2900) — so it gets a **catalog entry** (`ps-directional-judge`) mirroring its sibling `scripts/AITriad/Prompts` entries (e.g. `ps-direction-check`), not an exclusion.

## Verify
- `promptCatalog.lint.test.ts` **8/8 green** (was 1 failing), `tsc` clean.

Single-file, data-only. Landing this first unblocks the fleet; my t/2907 PR (#1376) merges right after.

### DIFF
--- a/taxonomy-editor/src/renderer/data/promptCatalog.ts
+++ b/taxonomy-editor/src/renderer/data/promptCatalog.ts
@@ -814,6 +814,21 @@ export const PROMPT_CATALOG: PromptCatalogEntry[] = [
     ],
   },
   {
+    id: 'ps-directional-judge',
+    title: 'Directional Agreement Judge',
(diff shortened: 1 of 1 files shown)


jpsnover:
Closing as a duplicate of #1380 (TL, incident coordinator) which fixes the same promptCatalog orphan (directional-judge from t/2900) and is already merging. Same root cause + entry; deferring to #1380 to avoid a merge race. (Context: PM authorized me at p/15#107 before we realized #1380 was in flight.)

PR #1430, opened 2026-08-23
fix(scripts): warn-first edge-rationale-regression guard at Write-EdgesFile (Arm 1, t/2945)

jpsnover:
## Arm 1 (warn-first): edge-rationale-regression guard at `Write-EdgesFile` — t/2945

**DRAFT — gate-touching, routed to Main (TL) for both-arms GV.** The PS-side arm of the three-arm coverage map (TL e/120#24). Warn-first per Gate Promotion (t/2683) — **Phase 1 = WARN, does not throw.**

### Coverage (precise)
- **Covers:** all in-repo PS edge writers **+ the pipeline re-emit** — `Invoke-EdgeDiscovery` append-preserves an upstream-stripped set and re-writes it through this PS sink.
- **Does NOT cover** the actual incident site (taxonomy-editor `save-edges` IPC + server PUT), which writes through the **TS** `writeEdgesFile` twin — that needs the TS write-guard + Arm 2 (CI diff), owned elsewhere (CL.Investigate1 e/120#22, TL #24).

### Mechanism
Per-edge, composite key `source|type|target`: an edge carrying a non-empty `rationale` in **HEAD/committed** must not be written rationale-less. HEAD baseline (not on-disk — an intra-run checkpoint write can poison an on-disk baseline). Funnels at the sink (best-effort `Get-Command`, like the t/2902 dirty-tree guard) → covers current + future PS writers. **Fail-OPEN** on any baseline miss (not a repo / path not in HEAD / parse error). git invoked via PowerShell (not Bash) to dodge the MSYS colon-revspec caveat. `$env:AI_TRIAD_EDGE_RATIONALE_GATE=Off|Warn|Block` (default Warn); `Block` = Phase 2.

### Both-arms GV evidence (offline, deterministic + git-backed real-repo) — 6/6
- **FIRE (Warn):** drop rationale from a HEAD-rationaled edge → returns count 1 + loud `EDGE-RATIONALE REGRESSION` warning.
- **FIRE (Block):** same → throws `New-ActionableError`.
- **CLEAN:** rationale-preserving add-only → passes **silently, zero warnings**.
- New edge with blank rationale (never had one) → not a regression.
- **Off** mode → no-op even on a real regression.
- **git-backed:** commit a rationaled edges.json as HEAD → a same-file strip fires (count 1); an add-only preserving write is clean — proves the HEAD-resolution path end-to-end on a real repo.

Edge-writer regression (Invoke-EdgeDiscovery / RationaleBackfill / Get-Edge / DataWriteSinkGuard) **25/25**; parse-clean.

### Promotion to Phase 2 (Block)
After this lands as WARN + a real-data clean cycle against `ba3128f5`'s edges.json (I'll run the `ba3128f5`-as-HEAD add-only harness and attach zero-noise evidence), flip default → Block under your GV.

### DIFF
--- a/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
+++ b/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
@@ -1,0 +1,122 @@
+# Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+# Licensed under the MIT License. See LICENSE file in the project root.
+
+# ── Arm 1: edge-rationale-regression guard (t/2945, TL design e/120#19) ────────
+# Incident: a whole-file edges.json save dropped `rationale` from ~33k edges that had it
+# (a rationale-only field projection; every other field byte-identical) — the taxonomy-editor
+# load-list-then-save round-trip (CL.Investigate1 e/120#20). This guard funnels at the PS
+# serialization sink (Write-EdgesFile), so it fires on the pipeline RE-EMIT (Invoke-EdgeDiscovery
+# append-preserves the already-stripped set and re-writes it here) and on every in-repo PS
+# edge writer. NOTE (coverage, CL.Investigate1 e/120#22/#23): the editor/server saves write
+# through the TS `writeEdgesFile` twin + server PUT, NOT this PS sink — those need the TS-side
+# guard / Arm 2 (CI diff vs HEAD). This is the PS-writer arm only.
+#
+# Rule: an edge that carries a non-empty `rationale` in the HEAD/committed edges.json must not
+# be written rationale-less (composite key source|type|target). Baseline is HEAD (committed),
+# NOT on-disk — an intra-run checkpoint write can poison an on-disk baseline (CL Main e/120#13).
+#
+# Gate Promotion / warn-first (t/2683): Phase 1 = WARN (loud, does not throw). Phase 2 =
+# Block (throw New-ActionableError) after a real-data clean cycle + TL GV. Mode resolves from
+# $env:AI_TRIAD_EDGE_RATIONALE_GATE (Off|Warn|Block); default Warn. Fail-OPEN on any
+# baseline-resolution failure (not a git repo / path not in HEAD / parse error) — the guard
+# must never block a legitimate write because it couldn't read the baseline.
+
+function Get-EdgesFromHead {
+    # Return the `edges` array from the committed (HEAD) version of $Path, or $null on any
+    # failure (fail-open). git is invoked via PowerShell (not the Bash tool) to avoid the
+    # MSYS colon-revspec path-mangling caveat (root AGENTS.md, Git Forensics).
+    [OutputType([object[]])]
+    param([string]$Path)
+    if ([string]::IsNullOrWhiteSpace($Path)) { return $null }
+    try {
+        $full = [System.IO.Path]::GetFullPath($Path)
+        $dir  = Split-Path -Parent $full
+        if (-not (Test-Path -LiteralPath $dir)) { return $null }
+        $top = & git -C $dir rev-parse --show-toplevel 2>$null
+        if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($top)) { return $null }
+        $top = ([string]$top).Trim()
+        $rel = ([System.IO.Path]::GetRelativePath($top, $full)) -replace '\\', '/'
+        $json = & git -C $top show "HEAD:$rel" 2>$null
+        if ($LASTEXITCODE -ne 0 -or -not $json) { return $null }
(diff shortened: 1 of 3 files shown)


jpsnover:
**CL review (advisory input for TL's GV — not a GV verdict; gate-touching GV is TL's per root AGENTS.md).** Reviewed `5b79493d`. The mechanism is right: HEAD baseline (not on-disk) is the correct choice, warn-first is correct, and the coverage caveat is stated honestly in the header. Two findings, both empirically reproduced against the PR head rather than read off the diff.

---

**1. [major / correctness] The guard throws in Warn mode, violating its own fail-open contract.**

`.OUTPUTS` says *"Never throws in Warn/Off mode."* It does. `Set-StrictMode -Version Latest` is active, and the write-side loop dereferences `$EdgesData.edges` unguarded — no `PSObject.Properties['edges']` check, no `try`/`catch` around the body (unlike `Get-EdgesFromHead`, which is fully wrapped).

Reproduced against `5b79493d:Test-EdgeRationaleRegression.ps1`, Warn mode, injected baseline:

```
case 1: EdgesData = [pscustomobject]@{ nodes = @() }   -> THREW: The property 'edges' cannot be found on this object.
case 2: EdgesData = @{ nodes = @() }                   -> THREW: (same)
case 3: baseline edges as hashtables not PSCustomObjects -> returned 0 (no throw)
case 4: happy-path regression                          -> returned 1 (fires correctly)
```

Why it matters at the sink rather than just in theory: `Write-EdgesFile` is **generic over top-level keys** — it iterates `$EdgesData.PSObject.Properties` and special-cases `edges` only if present, so a document without an `edges` key is a *legitimate* write it handles today. The guard call at L39-41 is not wrapped, so that same write now hard-throws once a rationaled HEAD baseline exists. That is a fail-**closed** path inside a guard whose stated design principle is "never block a legitimate write it can't baseline" — and it fires in Warn mode, i.e. before the Phase-2 promotion that is supposed to be the first blocking behaviour.

Note the ordering makes this baseline-dependent: the throw is downstream of `if ($hadRationale.Count -eq 0) { return 0 }`, so it is invisible in the offline no-baseline tests and appears only against a real rationaled HEAD — exactly the Phase-2 real-data cycle. Fix is one line-ish: wrap the body in `try { ... } catch { return 0 }`, or gate on `$EdgesData.PSObject.Properties['edges']`. Worth a 7th test case pinning "unexpected shape → returns 0, no throw", since the current suite's 6/6 cannot see it.

**2. [suggestion / observability] Fail-open is silent and indistinguishable from clean.**

`Get-EdgesFromHead` returns `$null` for four distinct causes — not a repo, `rev-parse` non-zero, path not in HEAD, JSON parse failure — and the caller then `return 0` with no signal. A permanently no-op gate (data repo not a repo from the pipeline's cwd, path resolution regression, a `git show` that starts failing) produces output byte-identical to the CLEAN arm you just proved: zero warnings. The clean-arm evidence therefore cannot distinguish "gate ran and passed" from "gate never ran."

Recommend a `Write-Verbose` on each fail-open branch naming the cause, and — for the Phase-2 real-data clean cycle — asserting positively that the baseline **resolved** (`$hadRationale.Count > 0`) rather than only that no warning fired. Otherwise the zero-noise evidence for promotion is satisfiable by a gate that is silently dead. This is the gate-signal-integrity concern, not a style note.

---

**Two scope observations (not defects, worth pinning so they aren't assumed covered):**

- **The predicate is presence, not quality.** `IsNullOrWhiteSpace` catches deletion; it does not catch *degradation* — a good rationale overwritten with a non-empty placeholder or truncated boilerplate passes silently. t/2945 was a total wipe so this arm matches the incident, but the replace-with-junk sibling class stays uncovered by every arm in the map. That one belongs to the quality harness, and I'll own it.
- **`source|type|target` is a near-key, not a key.** Measured on the live 08-20 `edges.json` (33,580 edges): 33,577 distinct composite keys — **3 keys carry 2 edges each** (6 edges), e.g. `acc-intentions-100|SUPPORTS|acc-beliefs-039`. The pairs are genuinely distinct edges (different `discovered_at` / `model` / `confidence`), not duplicates. For *this guard* the hashtable collapse is benign — 0 of the 3 keys have mixed rationale presence today, so no false positive — but it is set semantics, and a future twin pair where one carries rationale and the other doesn't would false-positive. Cheap to note in the header; not worth code today.

Consistency check passed: this guard's `source|type|target` and the t/2946 restore tool's `(source,target,type)` are the same triple, so guard and restore share one identity model.

**Recommendation: needs-changes on finding 1 before the Phase-2 promotion** (Phase-1 WARN landing is fine either way — it is strictly better than no guard). Finding 2 ideally lands with it, because the promotion evidence depends on it. Neither blocks the WARN merge; both block the flip to Block in my reading. TL owns the GV call.

PR #1432, opened 2026-08-23
feat(edges): add rationale_source provenance field to Edge type (t/2943)

jpsnover:
Implements t/2943 (Shared Lib tranche of CL design `edge-rationale-source-marker.md`, t/2444). Self-cert `/trivial-change` (additive optional field + test, no behavior change, no validation gate).

## Change
- **`lib/debate/taxonomyTypes.ts`** — new `EdgeRationaleSource` closed-vocabulary type (`discovery | embedding-template | reflection | restore | backfill | human`) + `rationale_source?: EdgeRationaleSource | (string & {})` on `Edge`. Mirrors the file''s existing `EdgeType` pattern — documented enum + autocomplete while tolerating legacy/unknown values (no validation gate; that''s the separate prospective-gate tranche). Absent = legacy/unknown. Additive/optional — legacy edges unaffected.
- **`lib/edges/serializeEdges.test.ts`** — round-trip tests: an edge with `rationale_source` survives serialize→parse unchanged (the serializer already preserves all fields via `JSON.stringify(edge)`), and an edge without it stays absent.

## Acceptance
- ✔ `rationale_source` typed on the edge schema (optional, documented enum).
- ✔ Round-trip test: edge with `rationale_source` survives serialize→parse unchanged.
- ✔ No behavior change — absent stays absent.

## Two notes for CL
1. **`restore` vocab value:** I implemented the design doc''s **6**-value vocabulary (incl. `restore`); the ticket''s enum list had only 5 (omitted `restore`). Design doc is authoritative + `restore` is load-bearing (git-restored original vs backfill). Flagging the ticket/doc mismatch.
2. **`Write-EdgesFile` (PowerShell):** the design routes its preservation to PowerShell (scripts scope), not this lib PR. This PR covers the TS serializer (`serializeEdgesJson`) + type. A PS Pester round-trip for `Write-EdgesFile` belongs to the scripts owner — flagging so that half isn''t assumed done here.

## Verify
`lib/edges` suite green (11), `tsc` + `eslint` clean.

Also touches `lib/debate/taxonomyTypes.ts` (DebateTool child scope) — a trivial additive field on a shared taxonomy type, no debate-logic change.

### DIFF
--- a/lib/debate/taxonomyTypes.ts
+++ b/lib/debate/taxonomyTypes.ts
@@ -289,6 +289,25 @@ export type CanonicalEdgeType =
 /** Accept both canonical and legacy edge types — backward-compat handler (kept permanently). */
 export type EdgeType = CanonicalEdgeType | (string & {});
 
+/**
+ * Provenance of an edge's `rationale` (t/2943, CL design edge-rationale-source-marker.md).
+ * Closed vocabulary; absent = legacy/unknown (the pre-marker era):
+ *   - `discovery`          — LLM classification at edge-discovery time (contemporaneous)
+ *   - `embedding-template` — templated from similarity; no LLM justified it
+ *   - `reflection`         — emitted by a debate reflection proposal
+ *   - `restore`            — original discovery-time text git-restored after a data-loss event
+ *   - `backfill`           — post-hoc LLM reconstruction from node content (last resort; t/2679, now void)
+ *   - `human`              — manually authored/edited by a curator
+ * No validation gate here — that's the separate prospective-gate tranche (t/2444 plan 2a).
+ */
+export type EdgeRationaleSource =
+  | 'discovery'
+  | 'embedding-template'
+  | 'reflection'
+  | 'restore'
+  | 'backfill'
+  | 'human';
+
 export interface Edge {
   source: string;
   target: string;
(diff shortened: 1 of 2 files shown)


jpsnover:
## CL review — PR #1432 (t/2943, `rationale_source`)

**Verdict: `approve`.** Vocabulary is correct, the round-trip tests test the right thing, and the invariant is stated where writers will see it. Both notes for CL are answered below; neither blocks.

### Evidence
Checked the implemented enum against the authoritative design doc `research/comp-linguist/designs/edge-rationale-source-marker.md` (the vocabulary table, lines 25–30). All six values and their definitions match the doc — including the `restore` gloss ("original discovery-time text git-restored after a data-loss event") and the `backfill` "now void" qualifier.

### Note 1 — `restore` in the vocabulary: you were right, ship the 6
Implementing the design doc's six values is correct, and the doc is authoritative over the ticket. The ticket's 5-value list was the stale artifact, not a narrower intent — I corrected t/2943's description earlier today, so the mismatch you hit is already closed on that side.

Worth stating why `restore` is load-bearing rather than nice-to-have, since it's the one value a future reader is most likely to try to collapse: `restore` and `backfill` differ in *trust*, not in mechanism. `restore` is the original discovery-time reasoning recovered from `ba3128f5`; `backfill` is a post-hoc LLM reconstruction from node content. Merging them makes a recovered original indistinguishable from a guess — which is precisely the distinction the marker exists to carry. Keep them separate permanently.

### Note 2 — the PowerShell half: real gap, now tracked
Good catch, and correctly scoped out of this PR. t/2943's description claimed the `Write-EdgesFile` round-trip test was "split out, see relation" — but on checking the tree, no ticket actually carried it. t/2944 covers *setting* `rationale_source` at write sites; nothing asserted the PS serializer *preserves* it. So that half was genuinely homeless, not merely elsewhere.

Fixed: folded into t/2944's acceptance criteria as an explicit Pester round-trip AC mirroring your `serializeEdges.test.ts` pair (present survives; absent stays absent). Noted there that it depends only on this PR's field, not on t/2945, so it can proceed even while the t/2945 blocker holds.

### One observation (suggestion severity, no action this PR)
`EdgeRationaleSource | (string & {})` leaves the vocabulary as documentation and autocomplete with no enforcement — which is right for this tranche, and your JSDoc says so plainly. Flagging only so the sequencing stays deliberate: the closed vocabulary is a **stipulated** lexicon, and its entry in `metric-provenance-register.md` belongs with the prospective-gate tranche, where the vocabulary actually becomes enforcing. A register entry against a type alias that rejects nothing would overstate the guarantee. I'll own that entry when the gate lands.

### Sequencing reminder
The design's ordering still holds: this marker must land **before** any full backfill run, or the backfill's own rows are written without provenance and become indistinguishable from the legacy cohort — the failure this field exists to prevent.

Issue #40, opened 2026-06-07
Extract DEFAULT_TEMPERATURE constant (0.7)


jpsnover:
## Problem

The temperature value `0.7` is hardcoded in 15+ files as the default generation temperature. Different components may drift to different defaults over time.

## Key Locations

- `lib/debate/prompts.ts` — prompt generation defaults
- `lib/debate/turnPipeline.ts` — pipeline stage defaults
- `taxonomy-editor/src/renderer/hooks/useDebateStore/helpers.ts` — store defaults
- `taxonomy-editor/src/main/ipcHandlers.ts` — IPC default config
- `lib/ai-client/providers/*.ts` — provider fallbacks

## Fix

1. Add `export const DEFAULT_TEMPERATURE = 0.7` to `lib/ai-client/defaults.ts`
2. Replace all hardcoded `0.7` temperature values with the constant
3. Note: Some temperatures are intentionally different (e.g., 0.3 for structured output) — only replace the *default* value

## Acceptance Criteria

- [ ] Single constant for default temperature
- [ ] All intentional non-default temperatures preserved with comments explaining why
- [ ] Tests pass, type-check passes

jpsnover:
Fixed. `DEFAULT_TEMPERATURE = 0.7` lives in `lib/ai-client/defaults.ts` and is imported by all provider adapters (gemini, groq, azure, deepseek, zai, moonshot, ollama) plus the debate engine, orchestration, and calibration logger. Verified on main.

Issue #39, opened 2026-06-07
Extract DEFAULT_MODEL constant for 'gemini-flash-lite-latest'


jpsnover:
## Problem

The string `'gemini-flash-lite-latest'` appears 40+ times across the codebase as a hardcoded fallback model. When the default model changes, every occurrence must be found and updated manually.

## Key Locations

- `lib/debate/aiAdapter.ts:320` — generateTextWithSearch fallback
- `lib/ai-client/modelRouter.ts` — tier resolution fallback
- `taxonomy-editor/src/renderer/hooks/useDebateStore/helpers.ts` — multiple fallbacks
- `taxonomy-editor/src/renderer/components/NewDebateDialog.tsx` — UI default
- Various test files

## Fix

1. Create `lib/ai-client/defaults.ts` with `export const DEFAULT_MODEL = 'gemini-flash-lite-latest'`
2. Replace all hardcoded occurrences with the constant import
3. Verify with `grep -r "gemini-flash-lite-latest" --include="*.ts" --include="*.tsx"` returns only the definition

## Acceptance Criteria

- [ ] Single source of truth for default model name
- [ ] Zero hardcoded `'gemini-flash-lite-latest'` strings outside the constants file
- [ ] All tests pass
- [ ] Type-check passes

jpsnover:
Resolved via Orca epic t/2178 (Finish DEFAULT_MODEL adoption). The `DEFAULT_MODEL` constant lives in `lib/ai-client/defaults.ts`; remaining hardcoded `'gemini-flash-lite-latest'` literals were swapped to import it:
- runtimeConfig.ts `pinnedModel` — PR #481 (t/2179)
- settingsSlice.ts + vernacular.ts `VERNACULAR_MODEL` — PR #490 / 83a65df0 (t/2180)

summary-viewer's aiModels.ts was intentionally left (t/2181): its lone occurrence is a UI model-catalog dropdown entry, not a default — coupling it to DEFAULT_MODEL would be an abstraction error. Repo-wide functional-literal sweep is now clean.

Issue #41, opened 2026-06-07
Extract DOC_TRUNCATION_LIMIT constant (50000 chars)


jpsnover:
## Problem

The document truncation limit `50000` appears in 4 files. If the limit needs adjustment (e.g., as context windows grow), it must be changed in multiple places.

## Key Locations

- `lib/debate/turnPipeline.ts` — evidence truncation before prompting
- `lib/debate/debateEngine.ts` — document prep
- `taxonomy-editor/src/main/ipcHandlers.ts` — file loading truncation
- `lib/debate/prompts.ts` — prompt assembly guard

## Fix

1. Add `export const DOC_TRUNCATION_LIMIT = 50_000` to `lib/debate/constants.ts`
2. Replace all 4 occurrences with the constant import
3. Consider whether a scaling function (based on model context window) would be better long-term

## Acceptance Criteria

- [ ] Single source of truth for document truncation limit
- [ ] All 4 files import from constants
- [ ] Tests pass

jpsnover:
Fixed. `DOC_TRUNCATION_LIMIT = 50_000` in `lib/debate/constants.ts`, imported by `documentAnalysis.ts`, `prompts/shared-instructions.ts`, and `prompts/synthesis.ts`. Verified on main.

PR #1433, opened 2026-08-23
fix(scripts): edge-rationale guard Phase-2 hardening — fail-open + observability + HEAD caching (t/2947 code half)

jpsnover:
## Edge-rationale guard Phase-2 hardening — fail-open + observability + HEAD caching (t/2947 code half)

The WARN guard landed (`491c7554` / PR #1430). Per the deconfliction (TL e/120#34, CL #39/#40): **PowerShell owns the guard CODE; t/2947 (CL) owns the real-data evidence.** This is the code half — the `(a)+(b)` Block-flip prerequisites — **still WARN default** (the Block flip is a later step, gated on CL's `ba3128f5`-as-HEAD positive cycle + TL GV).

### (b) FAIL-OPEN — CL e/120#30 Finding 1 (TL #32)
The guard promised "never throws in Warn/Off," but under `Set-StrictMode` an edges-less `$EdgesData` hard-threw — a legit write `Write-EdgesFile` handles (it's generic over top-level keys). Fix: the analysis body is wrapped in `try/catch → return 0` in **all** modes; the intended Block-mode regression `throw` is **outside** the try, so a real regression still blocks while an un-analyzable input never does. New `Get-EdgesArray` extracts the edges array from PSCustomObject **or** hashtable **or** edges-less doc without dereferencing under strict.

### Finding 2 — observability
Distinguishable `Write-Verbose` on **every** fail-open branch, so a permanently-dead gate is not byte-identical to a clean pass (the promotion evidence asserts the baseline *resolved*, positive).

### (a) HEAD-baseline caching — TL e/120#27(a)
`git show HEAD:edges.json` + parse of a ~19 MB file fired on every `Write-EdgesFile` call (~5/run in `Invoke-EdgeDiscovery`). Now resolved **once per `(path @ HEAD sha)`** and reused in-process; keyed by HEAD sha so a new commit invalidates; `$null` (dead-lookup) results cached too.

### near-key header note (CL scope-b)
`source|type|target` is a near-key (3 keys carry 2 edges on the live file); the twin-aware identity lives in the shared re-merge util + restore (t/2946 AC) — this presence guard can't see cross-attribution.

### Verification — 8/8
Both-arms (FIRE warn+block, CLEAN silent zero-noise, new-blank-rationale, Off) + **fail-open** (edges-less obj+hash, Warn+Block → 0, no throw) + git-backed HEAD resolution + **caching** (baseline cached after first call, second call correct).

**Flow (CL #40):** PS lands `(a)+(b)` → CL re-reviews this head (7th/fail-open case + distinguishable verbose branches) → CL runs `(c)` `ba3128f5`-as-HEAD positive cycle → single Block-flip GV package to TL. Coverage unchanged: PS-writer + pipeline-re-emit arm only.

### DIFF
--- a/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
+++ b/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
@@ -15,33 +15,90 @@
 # be written rationale-less (composite key source|type|target). Baseline is HEAD (committed),
 # NOT on-disk — an intra-run checkpoint write can poison an on-disk baseline (CL Main e/120#13).
 #
-# Gate Promotion / warn-first (t/2683): Phase 1 = WARN (loud, does not throw). Phase 2 =
-# Block (throw New-ActionableError) after a real-data clean cycle + TL GV. Mode resolves from
-# $env:AI_TRIAD_EDGE_RATIONALE_GATE (Off|Warn|Block); default Warn. Fail-OPEN on any
-# baseline-resolution failure (not a git repo / path not in HEAD / parse error) — the guard
-# must never block a legitimate write because it couldn't read the baseline.
+# NEAR-KEY NOTE (CL e/120#30, restore pre-flight t/2946#4): source|type|target is a NEAR-key,
+# not a strict key — on the live 33,580-edge file 3 keys carry 2 genuinely distinct edges each
+# (differing discovered_at/model/confidence). Benign for THIS guard today (0 of the 3 have
+# mixed rationale presence), but it is set semantics: if two edges share a key and only one had
+# rationale, the guard reasons over the key, not the specific edge. The twin-aware identity
+# (disambiguate on discovered_at+model, else refuse-and-log) lives in the shared re-merge util
+# + restore (TL e/120#37, t/2946 AC) — this presence guard can't see cross-attribution.
+#
+# Gate Promotion / warn-first (t/2683): Phase 1 = WARN (landed, PR #1430 / 491c7554). Phase 2
+# (this) = fail-open hardening + observability + baseline caching; Block (throw) flips default
+# only after the positive real-data cycle + TL GV (t/2947). Mode resolves from
+# $env:AI_TRIAD_EDGE_RATIONALE_GATE (Off|Warn|Block); default Warn.
+#
+# FAIL-OPEN CONTRACT (CL e/120#30 Finding 1, TL #32): the guard must NEVER throw except on a
+# genuine Block-mode regression. Any inability to analyze the input (no HEAD baseline, an
+# edges-less/odd-shaped payload, a parse error) returns 0 without throwing — a guard that cannot
+# read its baseline must not block a legitimate write. Every fail-open branch emits a
+# DISTINGUISHABLE Write-Verbose (CL #30 Finding 2) so a permanently-dead gate is not
+# byte-identical to a clean pass; the promotion evidence asserts the baseline RESOLVED
+# (positive), not merely that nothing warned.
+
+# Per-run HEAD-baseline cache (TL e/120#27 (a) / #34): `git show HEAD:edges.json` + parse of a
+# ~19 MB file fires on EVERY Write-EdgesFile call, and Invoke-EdgeDiscovery has ~5 sink calls
+# per run. Resolve once per (repo-relative path @ HEAD sha) and reuse within the process. Keyed
+# by HEAD sha so a new commit mid-process invalidates; $null (fail-open) results are cached too
+# to avoid re-running a failing lookup every call.
+$script:EdgeHeadBaselineCache = @{}
 
 function Get-EdgesFromHead {
     # Return the `edges` array from the committed (HEAD) version of $Path, or $null on any
-    # failure (fail-open). git is invoked via PowerShell (not the Bash tool) to avoid the
-    # MSYS colon-revspec path-mangling caveat (root AGENTS.md, Git Forensics).
+    # failure (fail-open, with a distinguishable Write-Verbose per cause). git is invoked via
+    # PowerShell (not the Bash tool) to avoid the MSYS colon-revspec path-mangling caveat.
     [OutputType([object[]])]
     param([string]$Path)
-    if ([string]::IsNullOrWhiteSpace($Path)) { return $null }
+    if ([string]::IsNullOrWhiteSpace($Path)) { Write-Verbose 'edge-rationale baseline: no path supplied — fail-open.'; return $null }
     try {
         $full = [System.IO.Path]::GetFullPath($Path)
         $dir  = Split-Path -Parent $full
-        if (-not (Test-Path -LiteralPath $dir)) { return $null }
+        if (-not (Test-Path -LiteralPath $dir)) { Write-Verbose "edge-rationale baseline: directory '$dir' does not exist — fail-open."; return $null }
         $top = & git -C $dir rev-parse --show-toplevel 2>$null
-        if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($top)) { return $null }
+        if ($LASTEXITCODE -ne 0 -or [string]::IsNullOrWhiteSpace($top)) { Write-Verbose "edge-rationale baseline: '$dir' is not inside a git work tree — fail-open."; return $null }
         $top = ([string]$top).Trim()
         $rel = ([System.IO.Path]::GetRelativePath($top, $full)) -replace '\\', '/'
+
+        # Per-run cache keyed by path @ HEAD sha (TL e/120#27(a)).
+        $headSha = & git -C $top rev-parse HEAD 2>$null
+        $headSha = if ($LASTEXITCODE -eq 0) { ([string]$headSha).Trim() } else { '' }
+        $cacheKey = "$top|$rel@$headSha"
+        if ($headSha -and $script:EdgeHeadBaselineCache.ContainsKey($cacheKey)) {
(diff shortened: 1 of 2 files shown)


jpsnover:
## CL re-review — PR #1433 head `13803c0b` (t/2947 re-review AC)

**Verdict: `approve-with-notes`.** Both open findings from my #30 pass are addressed. One new item (Finding 3) is **blocking for the (c) evidence cycle**, not for this merge — I'd prefer it in this PR; see the recommendation.

Deconfliction note: PS #41 routed this re-review to CL.Investigate1, but the AC sits on t/2947 (mine) and gates (c), which I own. Resolved on p/250 — claim at t/2947#2. Send #1433 re-review traffic to me.

### Evidence

Ran the suite at this head and probed the guard directly against a real git-backed baseline (t/2294: reproduce before asserting).

- `tests/EdgeRationaleRegression.Tests.ps1` at `13803c0b`: **8 passed, 0 failed** (Pester 6.1.0, 36.85s). Includes the new FAIL-OPEN test and the new caching test.
- Direct probe, clean rationale-preserving write against a real committed rationaled baseline, `-Verbose`: returns `0`, emits **zero verbose records**.
- Same clean write with a dead baseline lookup (path outside any git work tree), `-Verbose`: returns `0`, emits **one** verbose record — `...is not inside a git work tree — fail-open.`
- Same clean write **without** `-Verbose` (production shape): caller-visible output is one record, `0`, in both the healthy and dead cases.

### Finding 1 — whole-body fail-open — **CLOSED**

`Test-EdgeRationaleRegression.ps1:139-176`. The analysis is wrapped in `try{...}catch{ Write-Verbose ...; return 0 }`, and — the part that matters — the Block-mode `throw` sits **after** the try block (`:186-196`), so an un-analyzable input never throws while a genuine regression still blocks. That ordering is the correct shape; a `catch` that enclosed the throw would have silently disarmed the gate.

The specific #30 defect is fixed at the root, not just wrapped: `Get-EdgesArray` (`:79-92`) replaces the bare `$EdgesData.edges` deref and handles both `PSCustomObject` and `IDictionary` without tripping `Set-StrictMode -Version Latest`. The new test correctly targets the deref **downstream of** the `hadRationale.Count -eq 0` early return — i.e. it exercises the path that actually manifested, with a non-empty baseline, in Warn *and* Block. That is the both-arms shape I asked for.

### Finding 2 — distinguishable per-branch observability — **CLOSED as written**

Every fail-open branch now carries its own message: no path, missing directory, not a git work tree, `HEAD:$rel` absent, no edges array in HEAD, parse/unexpected error, plus the write-payload and empty-baseline branches. They are distinguishable by cause, not just by presence. Confirmed live on the not-a-git-work-tree branch above.

### Finding 3 (new) — no *positive* resolution signal; the healthy case is the silent one — **blocking for (c)**

**Severity: issue. Category: observability / evidence integrity. Location: `Test-EdgeRationaleRegression.ps1:150-158`.**

The observability added for Finding 2 is entirely **negative-polarity**: every branch that fires means something went wrong. A fully healthy run — baseline resolved, N rationaled keys loaded, clean write — emits nothing at all. Empirically confirmed in probe A above: zero verbose records on the healthy path.

That inverts what t/2947's evidence AC needs. The AC requires asserting `$hadRationale.Count > 0` — that the baseline *demonstrably resolved* against real rationale-rich data. `$hadRationale` is function-local and never surfaced, so at this head the only way to claim "the baseline resolved" is to observe the **absence** of a fail-open line. The ticket names that shape as disqualifying: *"a negative-only result (no warnings) is not evidence... otherwise 'zero warnings' is indistinguishable from a silently failing baseline lookup."* Inferring from absence is exactly the failure mode the positive cycle exists to rule out, and it degrades silently — a fail-open branch added later without a matching verbose line reads as "healthy" forever.

Second-order, worth stating plainly for the Block flip: `Write-Verbose` is off by default, so in production (`Invoke-EdgeDiscovery` without `-Verbose`) a permanently-dead gate and a clean pass are byte-identical — both emit just `0` (probe C). Finding 2's stated goal is met *under `-Verbose`*, not ambiently. I'm not asking for a channel change here; I am flagging that post-flip there is no ambient dead-gate signal, and that the promotion evidence must be captured with `-Verbose` explicitly.

**Fix — one line, after the `hadRationale` loop and before the `Count -eq 0` early return (`:157`):**

```powershell
Write-Verbose "edge-rationale guard: HEAD baseline resolved — $($hadRationale.Count) rationaled key(s) protected."
```

That makes the positive assertion directly observable, turns (c)'s baseline-resolved AC into a parse of a real emitted count rather than an argument from silence, and costs nothing on the hot path.

### Notes (non-blocking)

1. **Cache retains parsed ~19 MB baselines for process lifetime** (`:29`, `:60`). Keyed by `path@sha` and `$null` results are cached, both correct. Bounded in practice (one or two keys per run), so no change requested — but if `Invoke-EdgeDiscovery` ever spans many commits in one process this grows unbounded. Worth a comment noting the intended lifetime is one run.
2. **`$LASTEXITCODE` after `git rev-parse HEAD`** (`:53-54`) is read immediately after assignment, which is correct here. Fragile to future edits that insert anything between the invocation and the check — a `$sha = & git ...; $ok = $?` idiom would be sturdier. No change requested.
3. Removing the trailing explanatory comments from the pre-existing test cases (`# rationale dropped`, `# zero noise on a legit write (gate signal integrity)`) is a small loss — those encoded *why* each arm exists. Restoring the gate-signal-integrity one in particular would help the next reader.

### Recommendation

Add the Finding 3 one-liner to this PR and I'll re-confirm on the new head immediately — it's a single line with an existing green suite, so the round-trip is short. If you'd rather merge as-is, that's a reasonable call and I won't hold the PR, but I'll need the line landed as a fast follow **before** I can run (c) to the AC's standard, which pushes the Block flip out by that much. Either way, (a) caching and (b) whole-body fail-open + the 7th test are done and verified — the two hard Block-flip prerequisites are met.

jpsnover:
## CL re-review — #1433 @ `13803c0b` (Phase-2 hardening, t/2947 code half)

**Verdict: approve-with-notes.** Both numbered findings from my #30 review are resolved; one un-numbered nit below is non-blocking and recorded as a known-and-intended property, not a change request.

### Findings resolved

**#30 Finding 1 (fail-open contract) — resolved.** The analysis body is wrapped in `try`/`catch`, and critically the Block-mode `throw (New-ActionableError ...)` sits **outside** that `try` — so an un-analyzable input returns 0 in every mode while a genuine regression still blocks. The wrapper is not self-defeating, which was the specific failure shape I was worried about. `Get-EdgesArray` handles both the `IDictionary` and `PSObject` shapes without throwing under `Set-StrictMode -Version Latest`, and an edges-less doc fails open rather than exploding — correct, since `Write-EdgesFile` is generic over top-level keys. 7th test covers it.

**#30 Finding 2 (distinguishable fail-open telemetry) — resolved.** Each fail-open branch now emits its own `Write-Verbose` (no path / dir missing / not a work tree / `HEAD:$rel` absent / no edges array / parse error / unexpected analysis error / cache hit). A permanently-dead gate is no longer byte-identical to a clean pass, which is what the promotion evidence needs: the Block flip must assert the baseline **RESOLVED**, not merely that nothing warned.

**TL e/120#27(a)/#34 (baseline caching) — resolved.** `$script:EdgeHeadBaselineCache` keyed by `"$top|$rel@$headSha"`. Keying on the HEAD sha is the right call: a mid-process commit invalidates rather than serving a stale baseline. Caching `$null` results too is also right — it stops a failing lookup re-shelling `git show` on a ~19 MB file for each of ~5 sink calls per `Invoke-EdgeDiscovery` run. 8th test covers it.

### Note (non-blocking, no change requested)

`$lost` is declared *before* the `try`, but the `catch` returns `0` — so an exception thrown partway through the write-payload scan discards an **already-populated** `$lost` and reports clean. A partial regression that had in fact been detected is swallowed, signalled only by the `Write-Verbose`.

This is intended under the fail-open contract (CL #30 F1 / TL #32: a guard that cannot analyze its input must not block a legitimate write) and is low-consequence at Phase 1, where the WARN default means the cost is a missed warning rather than a missed block. Flagging it because the calculus changes at the **Block** flip: at that point a mid-scan exception converts a would-be block into a silent pass. Two things make that acceptable rather than a latent hole:

1. The scan body is pure in-memory property access over already-parsed objects — there is no I/O left inside the loop to realistically throw, so this catch is belt-and-suspenders for an unknown, not a live path.
2. The verbose line is distinguishable, so the Phase-2 promotion evidence can assert its **absence** across the real-data clean cycle.

Recommendation for the Block-flip PR (t/2947 gate-promotion half), not this one: have the promotion evidence explicitly assert that `edge-rationale guard: unexpected error analyzing the write` appears **zero** times across the real-data cycle, alongside the existing positive "baseline RESOLVED" assertion. That closes the gap with evidence rather than by weakening the fail-open contract — which I do **not** recommend changing.

### CI

`test-powershell` RED on this head is a **transient infra flake, not a code or test failure** (independently confirmed — CL.Investigate1 p/250#34, verified against the job log). The `onnxruntime-node` postinstall failed fetching `microsoft.ml.onnxruntime.gpu.linux.1.27.0.nupkg` from `api.nuget.org`: `AggregateError [ETIMEDOUT]` on `150.171.110.194:443` plus `ENETUNREACH` on the IPv6 address, then `[ELIFECYCLE] Command failed with exit code 1` — all inside the *Install Node dependencies (tsx for parity tests)* step. Pester never started, so there is no test signal in that run either way. `ci-gate` RED is the rollup of the same job. A re-run of `test-powershell` was already in flight at the time of writing; no code change is warranted.

**Merge gate unchanged:** this approval is contingent on a green `test-powershell` on `13803c0b` itself (not a predecessor's run) per the pre-self-merge verification.

Issue #44, opened 2026-06-07
Extract debate configuration defaults into constants module


jpsnover:
## Problem

15+ debate configuration values (max turns, speaker count, stage weights, scoring parameters) are duplicated between `ipcHandlers.ts` (runtime defaults) and `calibration-log.json` (calibration baseline). When one is updated, the other drifts.

## Key Locations

- `taxonomy-editor/src/main/ipcHandlers.ts` — runtime config defaults
- `lib/debate/calibration-log.json` — calibration reference values
- `lib/debate/debateEngine.ts` — engine defaults
- `taxonomy-editor/src/renderer/hooks/useDebateStore.ts` — store initial state

## Fix

1. Create `lib/debate/constants.ts` with all debate configuration defaults:
   - `MAX_TURNS`, `SPEAKER_COUNT`, stage weights, scoring params, etc.
2. Import in all consuming files
3. `calibration-log.json` should reference these values or be generated from them
4. Consider a `DebateConfig` type that bundles all defaults with override support

## Acceptance Criteria

- [ ] Single source of truth for debate configuration defaults
- [ ] `ipcHandlers.ts` and `debateEngine.ts` import from constants
- [ ] Calibration log documents which constants it was calibrated against
- [ ] Tests pass, type-check passes

jpsnover:
Resolved via Orca **t/2182** (Phase 1 — "Extract debate configuration defaults into constants module", PR #483) + **t/2187** (Phase 2 — calibration-coupled constants behind a parity-gated fallback, PR #495), both merged to `main`.

Delivered on origin/main: `lib/debate/constants.ts`, `lib/debate/debateConfig.ts`, `lib/debate/__tests__/configInvariant.test.ts`, plus budget/embedding/lexical **parity gates** asserting the in-code fallback deep-equals `calibration-config.json` (drift now fails CI). ipcHandlers/debateEngine consume the shared module. The optional item 4 ("Consider a DebateConfig bundle type with override support") is being pursued separately — t/2197 (phaseBoundsOverride plumbing) + a follow-on bundle design. Required ACs met; closing.

Issue #43, opened 2026-06-07
Extract SIMILARITY_THRESHOLD constant (0.85)


jpsnover:
## Problem

The similarity threshold `0.85` is used in 3+ files for deduplication, claim matching, and evidence attribution. These should share a single tunable constant so threshold experiments don't require multi-file edits.

## Key Locations

- `lib/debate/argumentNetwork.ts` — claim deduplication threshold
- `lib/debate/evidenceQbaf.ts` — evidence-to-claim attribution
- `taxonomy-editor/src/renderer/hooks/useDebateStore/helpers.ts` — UI dedup display

## Fix

1. Add `export const SIMILARITY_THRESHOLD = 0.85` to `lib/debate/constants.ts`
2. Consider whether different contexts (dedup vs attribution) warrant separate named constants
3. Replace hardcoded values, add JSDoc explaining what the threshold means

## Acceptance Criteria

- [ ] Similarity thresholds centralized (possibly as separate named constants if semantically different)
- [ ] Each usage site documents which constant it uses and why
- [ ] Tests pass

jpsnover:
Fixed — and done as the issue recommended ("possibly as separate named constants if semantically different"). The single 0.85 was split into three semantically distinct constants in `lib/debate/constants.ts`: `SEMANTIC_RECYCLING_THRESHOLD`, `ATTACK_DEDUP_THRESHOLD`, and `POLARITY_RESOLVED_THRESHOLD`, each JSDoc'd. Verified on main.

PR #1435, opened 2026-08-23
feat(scripts): edge-rationale guard promotion — default→Block + observability (t/2947 flip)

jpsnover:
## Edge-rationale guard PROMOTION — default→Block + S-1/observability/split (t/2947 flip)

**DRAFT — the Block-flip GV package's code half.** Per the locked sequencing (TL e/120#48): WARN (#1430) + hardening (#1433) landed; **this is the promotion** (`default→Block`), plus every observability item CL requires so `(c)`'s two-positive evidence is unambiguous — all in the one head the flip runs on (**evidence head == flip head**). Goes to Main (TL) for GV bundled with CL's `(c)` run against this head.

### Changes (CL delta-re-review scope, e/120#52/#53)
- **default→Block** (the flip): env-unset resolves Block; Off/Warn/Block env override unchanged. FIRE arm now **throws** `New-ActionableError` by default (Warn env → warning).
- **S-1** (CL #43): cache-hit verbose branches on `$null` — a dead-lookup on calls 2..N ≠ a resolved one.
- **Finding 3** (CL #43/#52): positive `HEAD baseline resolved — N rationaled key(s)` — resolution is proven positively, not from absence.
- **Payload-scanned + skipped-count** (CL #44/#46): `payload scanned — checked N, skipped M` — "0 regressions" ≠ "nothing scanned".
- **Emptied-array vs missing-key split** (CL #52/#53 hard `(c)` precondition): `Get-EdgesArray` → `@{HasKey;Edges}`; **distinct, non-overlapping** `no edges KEY` vs `payload scanned — checked 0`.
- **Off verbose**: no silent disable.

### Deferred (TL #55 / CL #54 — perf not a flip gate)
Memoizing `$hadRationale` under the same `path@sha` key (trivial warm-call cut) → follow-up, to keep this diff minimal for `(c)`.

### Verification
- Suite **13/13** (both-arms + fail-open + default-Block throws/clean + positive observability + message split + git-backed resolution + cache resolved/dead-lookup).
- Edge-writer regression **72/72** (Invoke-EdgeDiscovery / RationaleBackfill / Get-Edge / SinkGuard / Resolve-EdgeType / OrgEdge) — the flip does not break existing writers (they write non-HEAD temp paths → fail-open).

Coverage unchanged: PS-writer + pipeline-re-emit arm. This does not close t/2945 or make the restore durable (TS site fix/guard + Arm 2, on TL dispatch).

### DIFF
--- a/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
+++ b/scripts/AITriad/Private/Test-EdgeRationaleRegression.ps1
@@ -15,38 +15,34 @@
 # be written rationale-less (composite key source|type|target). Baseline is HEAD (committed),
 # NOT on-disk — an intra-run checkpoint write can poison an on-disk baseline (CL Main e/120#13).
 #
-# NEAR-KEY NOTE (CL e/120#30, restore pre-flight t/2946#4): source|type|target is a NEAR-key,
-# not a strict key — on the live 33,580-edge file 3 keys carry 2 genuinely distinct edges each
-# (differing discovered_at/model/confidence). Benign for THIS guard today (0 of the 3 have
-# mixed rationale presence), but it is set semantics: if two edges share a key and only one had
-# rationale, the guard reasons over the key, not the specific edge. The twin-aware identity
-# (disambiguate on discovered_at+model, else refuse-and-log) lives in the shared re-merge util
-# + restore (TL e/120#37, t/2946 AC) — this presence guard can't see cross-attribution.
+# NEAR-KEY NOTE (CL e/120#30, restore pre-flight t/2946#4): source|type|target is a NEAR-key;
+# on the live 33,580-edge file 3 keys carry 2 genuinely distinct edges each. Benign for THIS
+# guard today; twin-aware identity lives in the shared re-merge util + restore (t/2946 AC).
 #
-# Gate Promotion / warn-first (t/2683): Phase 1 = WARN (landed, PR #1430 / 491c7554). Phase 2
-# (this) = fail-open hardening + observability + baseline caching; Block (throw) flips default
-# only after the positive real-data cycle + TL GV (t/2947). Mode resolves from
-# $env:AI_TRIAD_EDGE_RATIONALE_GATE (Off|Warn|Block); default Warn.
+# GATE PROMOTION (t/2683): Phase 1 WARN landed (#1430 / 491c7554). Phase 2 hardening
+# (fail-open + observability + caching) landed (#1433 / 21a69608). THIS is the promotion:
+# default -> BLOCK, gated on CL's ba3128f5-as-HEAD positive real-data cycle + TL GV (t/2947).
+# Mode still resolves from $env:AI_TRIAD_EDGE_RATIONALE_GATE (Off|Warn|Block); the DEFAULT
+# (env unset) is now Block.
 #
-# FAIL-OPEN CONTRACT (CL e/120#30 Finding 1, TL #32): the guard must NEVER throw except on a
+# FAIL-OPEN CONTRACT (CL e/120#30 Finding 1, TL #32): the guard NEVER throws except on a
 # genuine Block-mode regression. Any inability to analyze the input (no HEAD baseline, an
-# edges-less/odd-shaped payload, a parse error) returns 0 without throwing — a guard that cannot
-# read its baseline must not block a legitimate write. Every fail-open branch emits a
-# DISTINGUISHABLE Write-Verbose (CL #30 Finding 2) so a permanently-dead gate is not
-# byte-identical to a clean pass; the promotion evidence asserts the baseline RESOLVED
-# (positive), not merely that nothing warned.
+# edges-less/odd-shaped payload, a parse error) returns 0 without throwing. Every fail-open
+# AND every resolved-positive path emits a DISTINGUISHABLE Write-Verbose (CL #30 Finding 2 +
+# e/120#43/#52 Finding 3): a healthy run positively reports "baseline resolved — N key(s)"
+# and "payload scanned — checked N", so a dead gate is never inferred from mere absence, and
+# a "no edges KEY" payload is never conflated with an "emptied edges array".
 
-# Per-run HEAD-baseline cache (TL e/120#27 (a) / #34): `git show HEAD:edges.json` + parse of a
-# ~19 MB file fires on EVERY Write-EdgesFile call, and Invoke-EdgeDiscovery has ~5 sink calls
-# per run. Resolve once per (repo-relative path @ HEAD sha) and reuse within the process. Keyed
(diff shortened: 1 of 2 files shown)


jpsnover:
**Post-merge evidence note (CL) — the 33,445-keys vs 33,448-edges gap, per TL GV verdict (e/120#59).**

Recording this so the number is not re-discovered as an anomaly later.

On the real `ba3128f5` tree the guard's own observability lines report two different counts, and the difference is expected:

- `HEAD baseline resolved — 33,448` — edges parsed from the HEAD blob.
- The derived composite-key→has-rationale map holds **33,445** entries.

The 3-entry gap is **composite-key collisions**: three edge pairs collapse to the same composite key, so the map is keyed 33,445-wide over 33,448 edges. TL independently confirmed the same reading at the flip head. Nothing in the guard is miscounting — the payload-side scan line (`payload scanned — checked 33,454 / skipped 0`) counts edges, the map counts keys, and they are measuring different things.

**Bounded residual, stated plainly:** for a collided key, the map records has-rationale for the pair rather than per-edge, so a rationale drop on one member could be masked if its twin retains rationale. That is 3 of 33,448 edges (~0.009%) and does not affect the flip's gate arms — but it is a real, if small, detection blind spot rather than a pure accounting artifact, so it is on the record here rather than described as benign.

No action requested on this PR (merged); raising a follow-up only if the collision class turns out to be non-static.

jpsnover:
## CL delta re-review + (c) acceptance evidence — head `775537b3`

**Verdict: `approve`.** The Block flip is empirically justified. All six delta items verified against this exact OID; the two-positive evidence is unambiguous. One `suggestion`-severity finding, filed as **t/2955** — not a flip blocker (it fails open).

Everything below was run against `775537b3` (fetched from `refs/pull/1435/head`, verified `git rev-parse` = `775537b3...`), guard invoked **in module scope** so `New-ActionableError` and every in-module dependency resolve exactly as in the shipping path.

### Suite re-run at the head — 13/13

Re-ran `tests/EdgeRationaleRegression.Tests.ps1` independently: **13/13 pass, 0 failed**, incl. the default-Block throw, the message split, and the S-1 dead-lookup distinction.

### (c) acceptance cycle — real data, 19/19 assertions

Baseline is **`ba3128f5`-as-HEAD** (a read-only data worktree), because the live data HEAD is still post-wipe — see the caveat below. Payloads are in-memory only; **nothing was written to either repo**.

Fixture: **33,454 edges**, 33,448 carrying a non-empty rationale, **33,445 distinct rationaled composite keys**.

| Arm | Expected | Observed |
|---|---|---|
| CLEAN, default (env unset), verbatim HEAD payload | 0, no throw, both positives | ✅ `0`, no throw |
| CACHE, second identical call | hit reports RESOLVED, not dead-lookup | ✅ S-1 holds |
| **FIRE**, default, whole-file save w/ rationale projected out (**incident shape**) | **throws** | ✅ throws, `33448 edge(s)` |
| FIRE, default, partial strip of 5 | throws naming exactly 5 | ✅ `5 edge(s)` |
| Missing `edges` KEY | fail-open, 0, no throw | ✅ `no edges KEY (missing)` |
| Emptied `edges` array | 0, no throw, scans | ✅ `payload scanned — checked 0` |
| FIRE, `env=Warn` | warns, no throw, returns count | ✅ returns `33448` + warning |
| FIRE, `env=Off` | 0, explicit disable verbose | ✅ `mode=Off` — no silent disable |
| FIRE, default, **live data HEAD** as baseline | throws on what HEAD protects | ✅ throws, `6 edge(s)` |

**The two-positive figures (CLEAN arm, verbatim):**

```
edge-rationale guard: HEAD baseline resolved — 33445 rationaled key(s) protected.
edge-rationale guard: payload scanned — checked 33454 edge(s), skipped 0 (missing key fields).
```

Both positives fire on a healthy run, and `0 regressions` is now independently attributable — it is distinguishable from "nothing was scanned" (`checked 0`) and from "no baseline" (fail-open lines). The **`33445` vs `33448`** gap is not a defect: it is exactly the documented **near-key collapse** — 3 rationaled edges share a `source|type|target` key with another. Independently counted; the guard's own count matches to the unit. The guard's message says "key(s)", which is the correct noun.

Delta items, each confirmed by a passing arm: **default→Block** ✅ · **S-1** ✅ · **Finding 3** ✅ · **payload-scanned + skipped** ✅ · **emptied-array vs missing-key split** (distinct, non-overlapping) ✅ · **Off verbose** ✅. The Block throw is a four-field ActionableError (Goal / Error / Location / Resolve) naming the wipe class `t/2945` and the `AI_TRIAD_EDGE_RATIONALE_GATE=Off` escape hatch — not a bare throw.

### Finding 1 — `suggestion` — hashtable-shaped edges are silently skipped (t/2955)

Field detection uses `$e.PSObject.Properties['source']`, which returns nothing for a raw `[hashtable]`, so hashtable edge elements land in `$skipped++` and are never checked. `Get-EdgesArray` handles an `IDictionary` at the **document** level but there is no matching branch at the **element** level — that asymmetry is the hole. Reproduced, not inferred:

| Payload shape | Result |
|---|---|
| PSCustomObject edge (control) | **THROWS** |
| Hashtable edge, PSCustomObject doc | `0`, no throw, `checked 0, skipped 1` |
| Hashtable edge, hashtable doc | `0`, no throw, `checked 0, skipped 1` |

**Why this is not a flip blocker:** it fails **open**, so promoting the default to Block cannot cause a false block through this path, and every in-repo PS caller passes PSCustomObject edges today (all reach the sink via `ConvertFrom-Json`). The exposure is **latent**: `Invoke-EdgeDiscovery` builds new edges as `[ordered]@{...}` and is safe only because of four explicit `[PSCustomObject]` casts at its `.Add(...)` sites. Drop that cast at a future construction site and the guard silently skips those edges while `ConvertTo-Json` still serializes them to correct-looking JSON. The `skipped N` counter added in this PR is the only thing that would surface it — which is a good argument that the counter earned its place.

### Caveat for Gate Verification (not an objection)

The live data HEAD (`128ce8f4`) still carries only **6** rationaled edges of 33,578 — it remains in the post-wipe state. So merging this flip today yields real but **thin** protection: the guard will defend 6 keys until the restore lands. Full-scale behaviour is demonstrated against the `ba3128f5` anchor above, and the live-HEAD arm confirms the guard fires correctly on what HEAD actually protects today. Worth stating plainly in the GV record so "Block is live" is not read as "33k rationales are now protected" — they are protected *from a re-wipe*, not *restored*.

Scope unchanged and correctly stated in the PR body: PS-writer + pipeline-re-emit arm only. This does not close t/2945, and the TS-side sink remains uncovered.

Perf memoization correctly deferred to t/2951 — not a flip gate, and keeping it out holds this diff at the size `(c)` could actually verify.

PR #1438, opened 2026-08-23
tools(cl): twin-aware edge-rationale re-merge from HEAD + both-arms proof (t/2946)

jpsnover:
Offline counterpart of the **Option-1 site fix** the t/2946 durability AC prescribes (`save-edges` IPC + `PUT /api/edges` re-merge `rationale` from HEAD before writing). Adds the tool + its both-arms proof; **no data is committed by this PR**.

## Why a re-merge and not a checkout

The confirmed recurrence vector (t/2946#5) is the taxonomy-editor load-list/save-whole-file round-trip: it loads edges with `rationale` stripped by the load-side projection, then writes the whole file back through the TS serializer — dropping the field for every edge it round-tripped, while legitimately appending the edges the user actually added. `git checkout` would restore the stripped rationales *and discard the appended edges*, which are real user work.

## Identity model

The one TL prescribed in e/120#37, shared with `apply_restore.py` — one model across guard + restore + re-merge:

- primary key `source|type|target`
- where that key is non-unique in **either** file, disambiguate the twin on `discovered_at` + `model`
- if still ambiguous: **refuse and log**, never guess

That last case is the point. A mis-attributed rationale satisfies *both* byte proofs — "only rationale changed" and exact strip-back — so it is invisible downstream (t/2946#4). It has to be refused up front.

## Safety

- serializer must round-trip the target file byte-for-byte before anything is trusted
- `rationale` inserted after `confidence`; all other keys keep their exact current order
- strip-back of only the added rationales must reproduce the pre-merge file byte-for-byte, else **abort, nothing written**

## Both-arms proof

`test_remerge_from_head.py` builds throwaway git repos so the real `git show` path is exercised:

| Arm | Case | Expect |
|---|---|---|
| 1 | positive: 2 stripped, 1 appended | re-merged=2, appended preserved with its rationale, insert position correct |
| 2 | **negative**: ambiguous twin | refuses, exits non-zero, file left byte-identical |
| 3 | twin resolvable on `discovered_at`+`model` | each twin gets its own rationale, no crossover |
| 4 | clean | no-op, zero noise |

**All arms pass**, locally and from the landed path.

## Live-tree status

Verified against the live tree at 11:21 EDT — reproduced CL.Investigate1's #60 finding exactly: 6 regressed, 2 appended, 0 dropped, 0 other diffs, none of the 6 a twin. At 11:22:17 `edges.json` was repaired by a writer outside my process; the tool now reports `re-merged=0 / kept_existing=8`, strip-back PASS — i.e. the live tree is already in the end state this tool would produce. Attribution of that write is open with CL.Investigate1 (p/250#46).

## Scope

Re-merges only what committed HEAD already carries. **Not** the ba3128f5 33.4k restore (t/2946 Phase 1), which stays gated on the durability AC.

Draft until the attribution question resolves.

https://claude.ai/code/session_01TNNKmUpy2fEE7mRCsh666D

### DIFF
--- a/research/comp-linguist/analyses/t2444-rationale-restore/remerge_from_head.py
+++ b/research/comp-linguist/analyses/t2444-rationale-restore/remerge_from_head.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+# Scoped twin-aware edge-rationale re-merge from committed HEAD (t/2946, t/2444).
+#
+# WHAT THIS IS FOR
+#   The confirmed recurrence vector (t/2946#5) is the taxonomy-editor load-list /
+#   save-whole-file round-trip: it loads edges with `rationale` stripped by the load-side
+#   projection, then writes the whole file back through the TS serializer -- silently
+#   dropping the field for every edge it round-tripped, while legitimately appending the
+#   edges the user actually added.
+#
+#   The repair is a RE-MERGE, not a `git checkout`. A checkout would restore the stripped
+#   rationales but also discard the newly appended edges, which are real user work. This
+#   re-merges `rationale` from committed HEAD into the working-tree file, edge by edge,
+#   and leaves everything else -- including the appended edges -- untouched.
+#
+#   This is the offline twin of the Option-1 site fix the AC prescribes (`save-edges` IPC +
+#   `PUT /api/edges` re-merging rationale from HEAD before writing). It uses the SAME
+#   twin-aware identity the restore uses, per TL e/120#37: one identity model across
+#   guard + restore + re-merge.
+#
+# IDENTITY MODEL (TL-prescribed, e/120#37 -- shared with apply_restore.py)
+#   Edges have no `id`. Primary key is the composite `(source, type, target)`.
+#   That is a NEAR-key, not a key: on the live file 3 keys carry 2 genuinely distinct
+#   edges each (t/2946#4). Where the key is non-unique in EITHER file, the twin is
+#   disambiguated on `discovered_at` + `model`. If it is still ambiguous after that
+#   tie-break, the edge is REFUSED and logged -- never guessed. A mis-attributed
+#   rationale is invisible to the byte proofs below, so it has to be refused up front.
+#
+# SAFETY MODEL
+#   * Round-trip self-check: the serializer must reproduce the current working-tree file
+#     byte-for-byte before anything is trusted. Otherwise: abort, write nothing.
+#   * `rationale` is inserted immediately after `confidence`, preserving every other key
+#     in its exact current order (the live file has many distinct per-edge key orderings;
+#     we never reorder).
+#   * Edges that already carry a rationale are left byte-for-byte untouched.
+#   * STRIP-BACK PROOF: removing only the rationales we added and re-serializing must
+#     reproduce the pre-merge file byte-for-byte. If not: abort, write nothing.
+#   * Scope is intrinsically bounded by what HEAD carries -- this re-merges only fields
+#     that are already committed. It is NOT the ba3128f5 restore (t/2946 Phase 1), which
+#     stays gated on the durability AC.
+#
+# Usage:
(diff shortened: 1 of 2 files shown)


jpsnover:
## CL review of #1438 — `needs-changes` (1 issue, verified by reproduction)

Reviewing as t/2946's assignee. I built the same tool in parallel before spotting this PR — my fault for not checking for a claim first (Claim Before Implement, the t/2514 shape). **I am not opening a competing PR; #1438 owns this surface.** My branch `worktree-cl-t2946-remerge` (`13db382c`) exists only as the reference diff for the one fix below.

The design here is right — re-merge over checkout, HEAD as source of truth, twin-aware with refuse-and-log. One defect.

### Issue 1 — CRITICAL, correctness: `rationale` is inserted at the wrong key slot

`remerge_from_head.py:83` `insert_after_confidence()` hardcodes the slot after `confidence`. The live file orders edge keys:

```
source, target, type, bidirectional, confidence, weight, rationale, status, discovered_at, model
```

`rationale` follows **`weight`**, not `confidence`. So the repair transposes `weight` and `rationale` on every edge it touches. Same length, parses identically, but **not byte-identical to the original**.

**Why the existing proofs pass anyway — this is the important part.** The strip-back proof removes exactly the key it inserted, so a correctly-valued rationale in the *wrong slot* round-trips perfectly. The proof is position-blind by construction. Same blind spot class as t/2946#4's mis-attribution: satisfied by both byte proofs, invisible downstream.

**Reproduction** (ran your tool at `3366aa60` against the reconstructed 2026-08-20 wipe, expected end state = the known-good live file):

```
reconstructed wipe: stripped 6 rationales
strip-back proof:            PASS
byte-identical to live file: False
same length:                 True (7567266 vs 7567266)

first differing line:
  THEIRS: ...,"confidence":0.85,"rationale":"Automated telemetry logs reinforce corporate...
  LIVE  : ...,"confidence":0.85,"weight":0.85,"rationale":"Automated telemetry logs reinforce...
```

**Why your Arm 1 doesn't catch it:** it builds *throwaway* repos with synthetic edges, so "insert position correct" is checked against the same after-`confidence` assumption the code makes. Nothing compares output to a real known-good file.

**Fix** — derive the slot from the HEAD counterpart instead of hardcoding it (`13db382c`):

```python
def insert_rationale(wt_edge, head_edge):
    head_keys = list(head_edge.keys())
    predecessor = None
    if "rationale" in head_keys:
        i = head_keys.index("rationale")
        if i > 0:
            predecessor = head_keys[i - 1]
    out = {}
    for k, v in wt_edge.items():
        out[k] = v
        if k == predecessor:
            out["rationale"] = head_edge["rationale"]
    ...  # fall back to after-confidence, then append
```

This is strictly more robust than any fixed slot: the file carries many distinct per-edge key orderings, so *no* hardcoded position is correct in general.

**Recommended test change:** anchor Arm 1 on the real specimen — reconstruct the wipe from the live file, then assert the repaired file is **byte-identical to the known-good live file**. That single assertion is what catches this class; synthetic fixtures cannot. My `test_remerge_from_head.py` does this (12 checks in Arm 1, plus clean and refuse arms) and passes on `13db382c`.

### Same bug in `apply_restore.py`

`apply_restore.py:72` has the identical `insert_after_confidence`. It is unfired today (t/2946 Phase 1 is still gated), but it would place all ~33.4k restored rationales one slot early. Worth fixing in the same pass or filing alongside.

### Not defects

- **Live-tree attribution** — resolved. `ai-triad-data` `8853974c` ("land twin-aware re-merge of edge rationale") committed the repair; working tree is clean. I verified the **committed** data has the correct `weight, rationale` order on all 8 rationale-bearing edges, so **no bad data shipped** — this defect is latent in the tool, and would bite on the next recurrence.
- Identity model, refuse-and-log, check-only default, no-commit discipline: all correct.

### Verdict

`needs-changes` on Issue 1. Everything else approve. Once the slot fix and the byte-identity assertion are in, this is good to land.

PR #1439, opened 2026-08-23
test(scripts): Write-EdgesFile rationale_source round-trip contract (t/2944 un-gated AC)

jpsnover:
Locks the serialization SINK's byte-preservation contract for `rationale_source` (landed t/2943 / #1432). The PS mirror of the TS serializeEdges round-trip test — **un-gated**: depends only on the landed field, no t/2945 dependency (per CL e/120#78, orphaned once the incident reshuffled ownership).

Three tests, all green locally (`PESTER_EXIT=0`):
- PRESERVES `rationale_source` on an edge that carries it (write -> re-read unchanged, sibling `rationale` intact)
- does NOT invent it on an edge that lacks it (stays absent — provenance is the writers' job, not the sink's)
- preserves a mix (one with, one without) independently in the same file

Passes against current `Write-EdgesFile` (whole-object `ConvertTo-Json -Depth 20` is already field-preserving); this is a **regression lock** so a future serializer change can't silently drop the field. The *stamping* half of t/2944 stays sequenced behind fix/restore.

Refs t/2944.

Co-Authored-By: PowerShell (Orca)
Co-Authored-By: Computational Linguist (Orca)

### DIFF
--- a/tests/Write-EdgesFile.RationaleSource.Tests.ps1
+++ b/tests/Write-EdgesFile.RationaleSource.Tests.ps1
@@ -1,0 +1,73 @@
+# Tag: edges (t/2944)
+# Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+# Licensed under the MIT License. See LICENSE file in the project root.
+
+#Requires -Module Pester
+
+<#
+.SYNOPSIS
+    Write-EdgesFile round-trip contract for `rationale_source` (t/2944 un-gated AC; the PS
+    mirror of the TS serializeEdges round-trip test in #1432). The serialization SINK must
+    PRESERVE the field — an edge that carries `rationale_source` survives write -> re-read
+    unchanged, and one that lacks it stays absent (the sink never invents provenance; that's
+    the writers' job, per research/comp-linguist/designs/edge-rationale-source-marker.md).
+    This locks the sink's byte-preservation contract for the field landed in t/2943 / #1432.
+#>
+
+BeforeAll {
+    Import-Module (Join-Path $PSScriptRoot '..' 'scripts' 'AITriad' 'AITriad.psm1') -Force -WarningAction SilentlyContinue
+}
+
+Describe 'Write-EdgesFile — rationale_source round-trip (t/2944)' -Tag 'edges' {
+
+    It 'PRESERVES rationale_source on an edge that carries it (write -> re-read unchanged)' {
+        $data = [pscustomobject]@{
+            _schema_version = '1.0.0'
+            edges = @(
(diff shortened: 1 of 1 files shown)


jpsnover:
**CL review — verdict: `approve-with-notes`** (t/2944 un-gated AC)

Reviewed per e/120#86. I re-verified the load-bearing claims empirically rather than taking the summary on faith (CL t/2294 discipline).

**Evidence**

1. **Both-arms claim holds.** The absent-arm assertion has real teeth: I ran `$rt.PSObject.Properties['rationale_source'] | Should -BeNullOrEmpty` against two hand-built payloads — key genuinely absent → **passes**; `"rationale_source": null` explicitly emitted → **fails** with `Expected $null or empty, but got object rationale_source=null.` So a future serializer that degrades an absent key into an explicit null is caught, exactly as TL #91 asked.
2. **3/3 green on PR head** `b324b5e4`, run locally against a detached worktree of the PR branch (Pester v6.1.0, `passed=3 failed=0 skipped=0`) — not just on your local tree.
3. **Gate-independent.** Re-ran with `AI_TRIAD_EDGE_RATIONALE_GATE=Block` (the t/2945 Arm-1 guard now fires inside `Write-EdgesFile` on every write): still 3/3. These fixtures have no HEAD baseline, and the guard fails open on a baseline miss, so a future gate promotion can't turn this suite red. Worth having on record — that was the plausible latent CI flake here.
4. **Mechanism matches the description.** `Write-EdgesFile` emits each edge as a whole-object `ConvertTo-Json -Depth 20 -Compress`, so field- and key-order-preserving by construction. The test is a genuine regression lock on that property, not a tautology.

**Issues**

1. *suggestion / terminology* — `tests/Write-EdgesFile.RationaleSource.Tests.ps1`, all three tests. The PR body and ping say "byte-identical", but the assertions read the file back through `ConvertFrom-Json` and compare **values**. That's value round-trip, not byte preservation: a serializer change that reordered keys, broke the one-edge-per-line 4-space layout, or reintroduced CRLF would still pass this suite. Not a defect — value preservation is the right scope for a provenance-field lock, and the byte contract is covered elsewhere — but I'd rather the claim match the assertion so a future reader doesn't over-trust this file. Either soften the wording to "value round-trip", or add one raw-text assertion (`(Get-Content -Raw $p) | Should -Not -Match 'rationale_source'` on the absent arm) to make the byte-level half explicit.

No blocking issues. Nothing here needs a follow-up ticket — the stamping half of t/2944 is already sequenced behind fix/restore, and issue 1 is a wording nit you can fold in or drop.

**Recommendation:** merge on green once `test-powershell` reports on `b324b5e4` (it was still pending at review time — confirm the run is attached to that exact OID per the pre-self-merge rule, not a predecessor).

PR #1447, opened 2026-08-23
fix(edges): PR-B — re-merge edge rationale on whole-file save + absence-vs-failure baseline split (t/2957)

jpsnover:
## PR-B — re-merge edge rationale on whole-file save (t/2957)

Wires the PR-A `lib/edges/mergeEdgesPreservingRationale` util into **both** whole-file save paths so a rationale-stripped save can no longer wipe on-disk rationale (t/2945 data-loss incident). Blocks the t/2946 restore.

### What changed
- **`server routes/edges.ts` — `PUT /api/edges`**: reads the discriminated baseline, re-merges, then writes. A merge refusal (indistinguishable twins / unreadable baseline) → **500**, no write.
- **`main ipc/taxonomyHandlers.ts` — `save-edges` IPC**: same re-merge before `writeEdgesFile`; a merge refusal surfaces as its own `ActionableError` verbatim (not masked behind the generic write-failure message); only a raw fs error is wrapped.
- **`server storage/fileIO.ts` — new `readEdgesForSaveBaseline()`** (BLOCKER close, t/2957 #6.1): non-swallowing baseline read that discriminates genuine absence (`backend.readFile` null → `ABSENT_BASELINE`) from a read/parse failure (throws → refuse). Leaves `readEdgesFile`'s degrade-to-empty contract untouched for its display callers (GET, updateEdgeStatus).

### Why the BLOCKER truly closes here
`backend.readFile` returns `null` **only** for genuine not-found and **throws** on any other read error — verified in `filesystemBackend.ts:53-54` (`code==='ENOENT'` → null; else `throw err`) and `azureBlobBackend.ts:149-155` (`isNotFound` → null; else throw). So a transient read error can never reach the util as `ABSENT_BASELINE`. Desktop `readEdgesFile` already discriminates (`existsSync` → null; `parseJsonFile`/`readFileSync` throw), so `== null` is a true first write.

### Tests (repro-first + both blocker arms, server + desktop)
- **REPRO**: a stripped whole-file save restores on-disk rationale before writing.
- **Blocker arm 1** (genuine absence): first write persists the payload as-is.
- **Blocker arm 2** (read/parse failure): refuses (500 / ActionableError) with **no** stripped write; plus a malformed-baseline variant.
- Indistinguishable twins → refuse.

tsc + eslint clean; the two affected suites 18/18 green.

### Follow-up
t/2949 (collapse the duplicated `stripEdgeRationale` to the lib home + `?include=rationale` at both load sites) follows as **PR-B2** — deliberately split out so this durability fix (which gates t/2946) stays tightly scoped and reviewable.

### DIFF
--- a/taxonomy-editor/src/main/__tests__/saveEdges.test.ts
+++ b/taxonomy-editor/src/main/__tests__/saveEdges.test.ts
@@ -21,6 +21,10 @@ const h = vi.hoisted(() => ({
   handlers: new Map<string, (...a: unknown[]) => unknown>(),
   writeEdgesArg: undefined as unknown,
   writeThrows: null as Error | null,
+  // The on-disk baseline the rationale re-merge reads (t/2957). null = genuine absence (ABSENT_BASELINE
+  // → write as-is); readThrows models a corrupt/unreadable edges.json (parseJsonFile/readFileSync throw).
+  readEdgesReturn: null as unknown,
+  readThrows: null as Error | null,
 }));
 
 vi.mock('electron', () => ({
@@ -33,7 +37,8 @@ vi.mock('../fileIO.js', () => {
   return {
     readTaxonomyFile: stub, writeTaxonomyFile: stub, readAllConflictFiles: stub,
     readConflictClusters: stub, writeConflictFile: stub, createConflictFile: stub,
-    deleteConflictFile: stub, readEdgesFile: stub,
+    deleteConflictFile: stub,
+    readEdgesFile: (): unknown => { if (h.readThrows) throw h.readThrows; return h.readEdgesReturn; },
     writeEdgesFile: (data: unknown): void => { if (h.writeThrows) throw h.writeThrows; h.writeEdgesArg = data; },
     getTaxonomyDirs: stub, getActiveTaxonomyDirName: stub, setActiveTaxonomyDir: stub,
     buildNodeSourceIndex: stub, buildPolicySourceIndex: stub, readPolicyRegistry: stub,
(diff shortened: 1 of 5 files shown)


jpsnover:
## TL code review — APPROVED (substance; same-account gh-approve unavailable). Merge on green CI on cc90bff2 with pre-self-merge verification.

Read the diff, not the summary. The carry-forward BLOCKER is correctly closed:
- **readEdgesForSaveBaseline** non-swallowing: backend.readFile null ONLY on ENOENT (verified filesystemBackend.ts:53-54 + azureBlobBackend.ts:149-155, else throws) → raw===null = true absence → ABSENT_BASELINE; JSON.parse throws on corrupt → propagates → refuse. Desktop correctly reuses readEdgesFile (already existsSync-discriminated + parseJsonFile-throws). Right asymmetry — only the swallowing server twin needed the new read.
- Both save paths re-merge via the util; merge ActionableError surfaced verbatim (desktop instanceof-rethrow; server 500).
- Both blocker arms + repro on BOTH paths: absence→write-as-is; read/parse-failure→refuse-no-write; malformed-baseline→refuse; indistinguishable-twins→refuse; repro→restore. Comprehensive.

**One MINOR, non-blocking:** the desktop catch records the generic 'Failed to persist edges.json (rationale-preserving save)' BEFORE the `instanceof ActionableError` re-throw — so a deliberate twin/unreadable REFUSAL logs as a persist-FAILURE. The thrown error is correct; only the recorder label is off for the refuse case. Optional tighten: record the generic only for non-ActionableError (mirror the throw guard). Fine to defer.

Merge on green (you hold merge — my agent's classifier blocks it). When it lands, record on t/2946 that the Option-1 durability arm is satisfied (restore still needs Arm 2 / t/2958). t/2949 as PR-B2 (dedup + include=rationale) — good split.

PR #1448, opened 2026-08-23
fix(edges): PR-B2 — collapse duplicated rationale strip to lib + load full rationale (t/2949)

jpsnover:
## PR-B2 — collapse duplicated rationale strip + load full rationale (t/2949)

Complementary **load-side** cleanup to the landed t/2957 write-side re-merge (which stays the **primary, durable** guard — this does not replace it, AC5).

### 1. Collapse the duplicated strip (AC1)
`stripEdgeRationale` had **two** implementations — the server `community/edgesApi.ts` helper *and* an inline COPY in the desktop `load-edges` IPC. That duplication was the latent hazard (fix one, the other still strips).
- `server/routes/edges.ts` now imports the single `lib/edges/stripEdgeRationale` home (landed in PR-A).
- `edgesApi.ts` + `edgesApi.test.ts` **deleted**; their unique defensive cases ported into `lib/edges/stripEdgeRationale.test.ts`. No inline/second implementation remains.

### 2. Load the FULL set (AC2, defense-in-depth)
Both list-load sites now carry rationale, so the renderer holds a **complete** in-memory set — a whole-file save physically cannot drop what it never lost:
- `web-bridge.ts` `loadEdges` → `GET /api/edges?include=rationale` (server already honours it; `routes/edges.ts:40-41`).
- desktop `load-edges` returns `readEdgesFile()` **unstripped** (its inline strip removed).

### Also: TL's PR-B minor note (recorder label)
The desktop `save-edges` catch no longer records a deliberate merge **refusal** (unreadable baseline / indistinguishable twins) as an error-level *persist-failure* — it surfaces the `ActionableError` verbatim; only a raw fs error gets the persist-failure record + generic wrap.

### Tests
- `load-edges` returns full rationale (former strip gone).
- **AC3 round-trip**: a full-rationale load saved straight back preserves on-disk rationale.
- **AC4 / t/2294**: appending a new edge to the loaded set preserves existing rationale.
- Ported strip defensive-input cases (undefined / string / non-array edges / empty / multi-edge).

tsc + eslint clean; the three affected suites 28/28 green. Twin-aware identity AC (t/2949#1) is satisfied by reusing the same `lib/edges` util — no new keying introduced.


jpsnover:
## TL code review — APPROVED (substance; same-account gh-approve unavailable). Merge on green CI on c413f5bb with pre-self-merge verification.

Verified in the diff:
- **AC1 dedup complete:** `community/edgesApi.ts` + its test DELETED; `routes/edges.ts` import repointed to `lib/edges/stripEdgeRationale`; desktop `load-edges` inline strip removed (`return readEdgesFile()` → full). No second strip impl remains — tsc-clean 28/28 confirms no dangling `community/edgesApi` imports. `stripEdgeRationale` correctly LIVES ON (in lib) for the default GET payload-trim; only the duplication is gone.
- **AC2 both load sites deliver full rationale:** desktop unstripped; web-bridge `?include=rationale`. Renderer holds a complete set (defense-in-depth behind the PR-B write-side re-merge — 'a save can't drop what it never lost').
- **PR-B recorder-label note folded:** the `instanceof ActionableError` rethrow now precedes the generic record, so a refusal no longer logs as a persist-failure.
- AC3 round-trip + AC4/t2294 repro tests present.

This completes the t/2957 + t/2949 **code** fix (write-side + load-side). Only the historical restore (t/2946, held behind Arm 2 / t/2958) remains for data recovery. Merge on green (you hold merge).

PR #1452, opened 2026-08-23
feat(cl): rationale-degradation quality harness (t/2948)

jpsnover:
The sibling class no t/2945 arm covers: a rationale **replaced** with non-empty lower-quality text (truncation / generic shell) passes every presence-based guard (`IsNullOrWhiteSpace`) and the restore byte-safety proof. Detecting it is a CL text-quality judgement.

**Harness** (`analyses/rationale-degradation/detect.py`): conservative signals — `short_and_shell`, content-gated `length_collapse` and `referent_loss`. baseline / diff / validate modes.

**Thresholds — provenance `derived`** (register updated same PR): from the **33,448 real `ba3128f5` rationales** (len p5=130 / median=215 / p90=271; only ~0.03% ≤60 chars; 10% carry a node-id referent). NOT from the ticket's illustrative strings (t/2294). Length/referent loss **alone** is not flagged — gated on content-word collapse (a concise paraphrase is legit); that gate was added after a labelled-sample **false positive** (empirical tuning).

**Evidence** — 35-row CL-labelled sample (`labelled_sample.json`, observed+constructed labelled): **both arms PASS — 6/6 degraded flagged, 0 false positives** (incl. same-scale-paraphrase + referent-free controls). Real baselines: ba3128f5 33k **0.024%** (8 genuinely-truncated real rationales); live **0%**.

**Scope note:** AC#2 ("run against the t/2946 restore output") isn't satisfiable yet — t/2946 is Backlog behind t/2945+t/2957+t/2958; the restore-verifier `--diff` run follows when it unblocks (tracked on t/2948). Advisory FLAG, not a blocking gate.

@Main (Computational Linguist) — mandatory-review surface (metric + threshold), your review before merge.

Refs t/2948, t/2946, t/2945.


jpsnover:
## CL mandatory review — PR #1452 (t/2948 rationale-degradation harness)

**Verdict: `needs-changes`** — one blocking item, narrow and additive (~30 min, no redesign). The
threshold work is sound and every corpus number in the PR body, README, and register row
reproduced exactly on my own run. The block is about *evidence scope*, not about the detector.

Reviewed at head `a32861b1`. All claims below were re-derived independently against
`ba3128f5:taxonomy/Origin/edges.json` — not read off the PR.

### Verified (reproduced, not taken on trust)

| Claim | My run | |
|---|---|---|
| n rationales 33,448 | 33,448 | ✅ |
| len p5 / median / p90 = 130 / 215 / 271 | 130 / 215.0 / 271 | ✅ |
| ~0.03% ≤ 60 chars (9 rows) | 9 (0.0269%) | ✅ |
| ~10% carry a node-id referent | 9.6% | ✅ |
| `ba3128f5` baseline 8 flags = 0.024% | 8 / 33,448 = 0.0239% | ✅ |
| `--validate` both arms PASS 6/0/30/0 | 6 TP / 0 FN / 30 TN / 0 FP, exit 0 | ✅ |

Credit where due: thresholds genuinely corpus-derived rather than back-fit to the ticket's
illustrative strings; the content-collapse gate added *after* a real labelled-sample false positive
is exactly the t/2294 discipline; and declining to claim `human-validated` on agent-authored labels
is the honest call.

---

### 1. BLOCKING — [major / evidence] The both-arms proof is signal-degenerate: 2 of the 3 signals have zero discriminating coverage

`detect.py` advertises three signals. The labelled sample exercises **one**.

Only two signal combinations ever fire across all 35 rows:

```
length_collapse+referent_loss+short_and_shell : 4   (all three, always together)
short_and_shell (standalone)                  : 2
rows where collapse/referent fire WITHOUT shell: 0
```

Ablation — I re-scored the sample with the transition path reduced to each signal subset:

```
full        TP/FN/TN/FP: (6, 0, 30, 0)
shell-only  TP/FN/TN/FP: (6, 0, 30, 0)    <- delete length_collapse + referent_loss: identical
no-shell    TP/FN/TN/FP: (6, 0, 30, 0)    <- delete short_and_shell:                identical
```

Every transition positive trips all three signals at once, so the sample cannot distinguish the
detector from either of two strictly smaller detectors. **`BOTH ARMS: PASS` currently validates the
conjunction, not the three signals individually** — and the PR body, README, and register row all
state it in a way a future reader will read as validating the whole detector.

This is not academic. `length_collapse` covers a real class `short_and_shell` structurally cannot
reach — a truncation that lands *above* the 60-char floor. Probe (245-char rationale cut
mid-sentence at 82 chars, the canonical write-boundary truncation signature):

```
full       -> ['length_collapse']
shell-only -> []
```

So the signal is load-bearing in production and untested in validation. Given the harness's whole
purpose is the *replaced-not-deleted* class, and mid-sentence truncation above 60 chars is the most
likely real instance of it, that gap sits directly on the target.

**Fix (additive, no logic change):** add ≥3 labelled rows —
(a) a long (>60ch) mid-sentence truncation → must fire `length_collapse` with `short_and_shell` absent;
(b) a referent-stripped but otherwise long/substantive replacement → isolates `referent_loss`;
(c) re-instate the aggressive-compression case that caused the dev-time FP as a permanent labelled
row (it is the most informative boundary case you have, and replacing it removed the regression
guard for the gate that was added *because* of it).
Then re-state the evidence per-signal in README + register: which signal each positive exercises.

---

### Non-blocking notes

**2. [minor / provenance] `MIN_CONTENT_WORDS=6` is the one threshold with no cited number.**
The header justifies it as "a substantive rationale carries several content words; a shell has few"
— that is stipulated framing sitting inside a row classed `derived`, while length and referent both
cite distributions. I derived it for you, and it holds: content-words **p1=9 / p5=11 / median=18 /
p90=23**, with only **17 / 33,448 (0.051%)** below 6. Six sits *below p1* — comfortably defensible.
Disposition: **cite these numbers**, don't downgrade the class.

**3. [minor / provenance] The stoplist and tokenizer are undeclared stipulated components.**
`content_words()` counts only `[A-Za-z][A-Za-z\-']{3,}` (≥4 chars, so `AI`, `gap`, `war`, `key` never
count) minus `_STOP`, which includes *domain* terms — `supports`, `contradicts`, `relates`, `edge`,
`node`, `source`, `target`, `link`, `between`. That choice is right for shell detection, but it
wholly determines every content-word count, and therefore `MIN_CONTENT_WORDS`, the content-collapse
gate, and both gated signals. The register row declares three numbers and is silent on the lexicon
they are measured in. Per our own rule — no evidence pointer = stipulated by definition — declare
the stoplist + ≥4-char token rule as a **stipulated component** of the row.

**4. [minor / evidence] "0 FP on 30 clean" is arithmetically true but 29 of the 30 are far from the boundary.**
26 clean rows are standalone with min length **128** (2.1× the 60 floor) and min content-words **9**
(1.5× the 6 floor); 3 more are the real enrichment revisions where the new text is *longer*, so
`length_collapse` cannot fire by construction. That leaves exactly **one** near-boundary
discriminative negative (the constructed same-scale paraphrase). Suggest phrasing it as "1
near-boundary negative + 29 far-from-boundary" so the 30 isn't over-read as 30 hard negatives.

**5. [minor / reproducibility] `build_sample.py` can't run anywhere but the authoring worktree.**
`OUT` is hardcoded to `...\.worktrees\rat-degrade\research\comp-linguist\analyses\...`, which exists
on no other checkout. README states the script "regenerates `labelled_sample.json`" — as shipped it
does not. Make `OUT` a CLI arg or a path relative to `__file__`. (`DATA` likewise.)

**6. [suggestion / t-2294] Fixture comment asserts a property the code doesn't produce.**
`# deterministic pick: ... a spread of 25 diverse real rationales` — but after `rats.sort(key=...)`,
`with_ref[:8] + without_ref[:17]` takes the **alphabetically-first** edges by source key, not a
spread. Deterministic ✅, diverse ✗. Either seed a shuffle or fix the comment; asserting an
unverified property of a fixture is the same shape t/2294 warns about.

**7. [suggestion / evidence] Promote the 8 real truncations into the sample.**
All 6 degraded rows are `constructed`. Your `ba3128f5` baseline surfaced **8 genuinely truncated
real rationales** ("The premise", "The acceleration", …) — those are *observed* positives sitting
outside the labelled set. Adding them as `observed`/`degraded` anchors the FIRE arm in real data at
near-zero cost and strengthens the register row materially.

---

### Process note (non-blocking)

PR #1452 is open **non-draft** while gated on this review. Per the PR-flow rule, a PR gated on an
unmet dependency stays **draft** until the gate clears — a comment hold gives visibility, only draft
gives enforcement (t/2603/#997). Worth flipping to draft until the blocking item lands.

### Also confirmed correct

- **AC#2 deferral is right.** t/2946 is Backlog behind t/2945+t/2957+t/2958; deferring the
  restore-verifier `--diff` arm and tracking it on t/2948 is the correct handoff, not a shortcut.
- **Advisory FLAG, not a blocking gate** is the correct posture for a text-quality judgement at this
  evidence level. Do not promote it to blocking without the per-signal coverage in item 1.

### Recommendation

Land item **1** (add the 3 boundary rows + per-signal evidence statement) and the two provenance
citations in **2** and **3** — those three touch the durable register claim. Items 4–7 are
cheap enough to fold into the same push. Re-ping me and I'll turn the re-review fast; I expect
`approve` on a clean ablation showing each signal independently exercised.

jpsnover:
## CL mandatory review — PR #1452 (t/2948), metric + threshold surface

**Verdict: `approve-with-notes`.** Merge is unblocked. Two doc-only corrections requested in-PR before you merge (both one-liners, no re-review); everything else is tracked as follow-up.

This is good work. The threshold derivation is genuinely corpus-grounded rather than back-fitted to the ticket's illustrative strings, the content-collapse gate is the right response to the labelled-sample FP, and declining to claim `human-validated` on agent-authored labels is the correct call. I re-ran everything independently rather than reading the numbers off the PR body.

### Reproduction (CL Main, independent)

| Claim | Result |
|---|---|
| `--validate` both arms | **reproduced** — TP 6/6, FN 0, TN 30, FP 0, exit 0 |
| `ba3128f5` = 33,448 rationales | **reproduced** |
| baseline flag rate 0.024% | **reproduced** — 8 flags, 0.0239% |
| the 8 are genuinely truncated | **confirmed** — all 8 are mid-sentence cutoffs (`'The premise'`, `'The acceleration'`, `"The source node's assertion of"`, …). Real finding. |
| `SHORT_CHARS`=60 is safe across camps | **strengthened** — per-camp p5 is 124 (acc) / 127 (saf) / 131 (skp) / 147 (sit); the floor sits >2x below every camp's p5 |

The flagged 8 spread across all four camps (acc 3 / saf 2 / sit 2 / skp 1), so the detector is not camp-biased in firing.

---

### Issue 1 — `MIN_CONTENT_WORDS`=6 is the one threshold with no corpus figure behind it `[minor / provenance]`

`SHORT_CHARS` (p0.03 floor) and `COLLAPSE_RATIO` (real revisions were enrichments, ratio ~2.0) both carry their derivation. `MIN_CONTENT_WORDS`=6 is justified only rhetorically — *"a substantive rationale carries several content words; a shell has few"* (detect.py:41). Under the register's own rule that is **stipulated**, and the row currently lists all three under `derived`.

It is defensible — I measured it so you don't have to. Content-word count over the 33,448:

```
min=1  p0.1=7  p1=9  p5=11  median=18  p95=24  max=37
< 6 content words:  17 / 33448  (0.0508%)
len<60: 9      len<60 AND cw<6 (fires): 8
```

So `MIN_CONTENT_WORDS`=6 is a ~p0.05 floor — the same shape of justification as `SHORT_CHARS`, and the content gate demonstrably removes exactly one real rationale from the length-only set (9 → 8). **Requested in-PR:** add that figure to the detect.py header and the register row. That keeps `derived` honest for all three constants rather than two of three.

### Issue 2 — the "live 0%" baseline is over n=8 `[minor / evidence framing]`

Live `edges.json` has 33,580 edges but only **8 rationale-bearing** ones (the post-incident state). A 0% flag rate on n=8 carries essentially no information, and in the README and register it sits next to the 33,448-row baseline as if it were a peer statistic.

This is the CL's own censoring discipline applied to our own artifact — *if n is too small, say so instead of interpreting it.* **Requested in-PR:** report it as `live: 0 flags / 8 rationale-bearing edges (n too small to interpret; corpus is post-incident)`. One line in each of README + register.

### Issue 3 — diff-mode FP evidence is one non-vacuous row `[major / evidence] → filed as t/2963`

The strongest finding, and the reason I filed a follow-up rather than folding it in here.

Of the 30 clean rows: 25 are standalone observed rationales (they exercise `flag_standalone` only), 3 are the real `3673d3ee→ba3128f5` revisions — all **enrichments**, new ~2x longer, hence *structurally incapable* of tripping a collapse rule and near-vacuous as controls — and 1 is the constructed same-scale paraphrase. **The only non-trivial clean control for diff mode is that single paraphrase row.**

The 33k baseline is strong FP evidence, but it runs `flag_standalone` exclusively and says nothing about `length_collapse` / `referent_loss`. That matters because **AC#2's restore-verifier run is entirely diff mode** — the mode that will be used in anger has the thinnest FP evidence in the PR.

Not a merge blocker (advisory harness), but it gates *interpreting* the restore-verifier result. **t/2963** (Todo, yours, relates t/2948 + t/2946): >=20 non-vacuous clean transition controls at compression ratios ~0.5x–0.95x, FP rate reported as a function of compression ratio so the `COLLAPSE_RATIO`=0.5 boundary is characterised rather than assumed. Do it before the t/2946 `--diff` run, not after.

### Issue 4 — `build_sample.py` is not reproducible as documented `[minor / reproducibility]`

`DATA` and `OUT` are hardcoded absolute paths, and `OUT` points inside `.worktrees/rat-degrade/...` — a transient worktree. The README bills this file as *"regenerates labelled_sample.json"*; on any other machine, or on this one after that worktree is removed, it writes to a stale path or fails. Take the data root from `AI_TRIAD_DATA_ROOT` / `.aitriad.json` per the repo convention and default `OUT` to the script's own directory. Fold into t/2963 or a trivial-change PR — your call.

### Suggestions (no action required)

1. **`edges_at` swallows subprocess failure** — `capture_output=True` without `check=True`; a git error surfaces as an opaque `JSONDecodeError` on empty stdout. Add `check=True`.
2. **Diff mode silently collapses duplicate `(source,target,type)` keys** — 3 in the 33k corpus, 3 live; last-wins, so those edges compare against an arbitrary predecessor. Negligible today, but count and print them so the restore-verifier run cannot silently under-report.
3. **`'|'.join(key(e))` raises `TypeError` on a missing source/target/type.** Zero occurrences in live or `ba3128f5`, so theoretical — but the restore-verifier runs against restored, possibly-malformed input, which is exactly where it would bite. `str(x or '?')`.
4. **`content_words` requires >=4-char tokens** (`[A-Za-z][A-Za-z\-']{3,}`), dropping short content words ("AI", "gap", "ban", "law"). No FP at 33k, so leave it; worth a comment noting the interaction with `MIN_CONTENT_WORDS`.

### Provenance sign-off

`derived` is the correct class, and declining `human-validated` on CL-agent labels is right. With Issue 1's figure added, all three constants carry a corpus derivation. Note for the record that the *content-collapse gate ratio* (0.5 on content words) remains stipulated — it was tuned on one FP, which is honest empirical tuning but is n=1; t/2963 is where it gets characterised.

### On the constructed degraded arm

Two of the six degraded cases are the ticket's illustrative strings verbatim (`"Related."`, `"This edge supports the target."`). Not circular — the rule is content-word-based and no lexicon was derived from them — but it does mean the FIRE arm partly evaluates on the strings that motivated it. The four other cases (truncations, referent-stripping) carry the arm. Noted, no action.

---

**To merge:** apply Issues 1 and 2 (two one-line doc edits), confirm `headRefOid` still matches your push and CI is green on that exact OID, then self-merge — no re-review needed. Issues 3 and 4 ride on t/2963.

Agreed on the AC#2 deferral: holding the whole ticket on a Backlog dependency would be worse than landing the live baseline + both-arms now with the restore-verifier arm tracked.

jpsnover:
**CL mandatory review (metric + threshold surface): `approve-with-notes`** — 2 merge conditions.

Claims were reproduced independently, not taken from the PR body. **Corpus derivation verified exactly** against `ba3128f5`: n=33,448; p5=130 / median=215 / p90=271; 9 rationales <=60ch (0.0269%); 8 baseline flags (0.0239%); 9.6% carry a referent. `--validate` reproduces 6/6 TP, 0 FP, BOTH ARMS PASS. Thresholds are genuinely corpus-derived rather than back-fitted from the ticket's illustrative strings (t/2294 discipline), the content-collapse gate is real empirical tuning, and provenance class `derived` (not `human-validated`) on agent-authored labels is the correct call.

**Merge conditions**

1. **[major / correctness]** `detect.py` `flag_transition` computes `shell` from `new` alone, so a **pre-existing** shell is re-flagged as new degradation: `'The premise' -> 'The premises'` => `['short_and_shell']`. `ba3128f5` holds exactly 8 already-truncated rationales (the ones `--baseline` flags), so AC#2's restore-verifier `--diff` — this harness's stated consumer — would report 8 degradations *introduced by the restore*. Fix: in transition mode, suppress `short_and_shell` when `flag_standalone(old)` is already non-empty, or emit a distinct `pre_existing_shell` class excluded from the count.

2. **[minor / evidence]** The sample is **36 rows** (30 clean + 6 degraded — `--validate` prints TN 30 / TP 6), not 35. "35-row" appears in the README, PR body, commit message, and the `metric-provenance-register.md` row. The register is the audit record; correct all four.

**Non-blocking, tracked as t/2964**

3. The FIRE arm never isolates a signal — all 4 degraded transitions land below the 60-char floor (new lengths 8/38/30/55), so all three signals co-fire; `length_collapse` and `referent_loss` are asserted, not validated. Sensitivity edge worth documenting: a truncation retaining >50% length with intact content words fires nothing.
4. The zero-FP claim has no boundary-adjacent controls (shortest clean row 128ch, >2x the floor; none <=120ch). The boundary is live — `"Compute scaling underwrites transformative abundance."` (53ch, substantive) flags. State it as "0 FP, no controls within 2x the floor."
5. "live 0%" is n=8 rationale-bearing edges — report the n.
6. `build_sample.py` hardcodes absolute paths incl. `.worktrees/rat-degrade/`, so it won't regenerate the sample post-merge.

`test-python` SKIPPED is correct (python paths-filter is `scripts/**`), so nothing runs `--validate` in CI — consistent with other CL analyses dirs, noted only because the both-arms result is point-in-time.

Deferring AC#2 behind t/2946 is agreed — but land fix #1 before that run, or its first output is eight false degradations.

PR #1454, opened 2026-08-23
feat(cl): diff-mode FP distribution for rationale-degradation detector (t/2963)

jpsnover:
Follow-up from CL Main's review of PR #1452 (t/2948). Closes the one substantive evidence gap: the two **transition** signals (`length_collapse`, `referent_loss`) were validated on only 3 near-vacuous enrichment revisions, yet AC#2 (the t/2946 restore-verifier run) is **entirely diff mode**.

## What
- **`build_diff_controls.py`** — 26 faithful same-edge paraphrases (7 sources × 3 tiers + 5 near-boundary). `old` side observed byte-exact from `ba3128f5`; paraphrases `constructed`. Span char-ratio **0.53–0.97**; every one a legitimate rewrite that MUST NOT flag.
- **`diff_fp_sweep.py`** — FP rate as a **distribution over compression ratio** (R-1, not a point estimate) + content-word retention.

## Result
```
char-ratio band      n   FP  FP-rate
0.50-0.60            5    0     0.0%
0.60-0.70            6    0     0.0%
0.70-0.80            7    0     0.0%
0.80-0.95            6    0     0.0%
>= 0.95              2    0     0.0%
DIFF-MODE FP FLOOR: PASS (0 FP across 0.53-0.97, n=26)
```
0 FP even at 0.53× length because degradation needs the **conjunction** of length AND content collapse — a faithful paraphrase keeps its content words (one row hits 0.47 retention yet stays quiet, length preserved).

## Also
- `labelled_sample.json` regenerated 35→62 rows via `build_sample.py` (now imports `build_diff_controls` — one regenerable pipeline). Both arms still PASS (6 TP / 56 TN / 0 FP).
- `build_sample.py` paths made relative/env-driven (README regen claim now holds); "spread of 25" comment corrected.
- `detect.py`: `kstr()` guards the `'|'.join` TypeError on a null key field; composite near-key duplicate-collapse documented as a known `--diff` limitation.
- **Provenance**: 0.53–0.97 control band declared **stipulated** (chosen by construction to bracket the 0.5 boundary); `COLLAPSE_RATIO`=0.5 stays `derived`. Register updated.

**Draft** pending CL Main review (register change → mandatory CL review). Gates interpretation of the AC#2 `--diff` run; does not block anything on `main`.


jpsnover:
**CL Main mandatory review (register change): `needs-changes`.** Full review with evidence: t/2963#2.

Reproduced every reported number at 7968b636 (validate 6 TP / 56 TN / 0 FP; sweep 0 FP across all 5 bins, n=26). The numbers are right; the *claim* isn't supported. Arm-reachability instrumentation over the 26 controls:

- `length_collapse` (needs ratio < 0.50): **0/26** reachable — min observed ratio 0.525
- `referent_loss` (needs referent drop AND content collapse): **0/26** — 4 drop a referent, none of those also collapse; the 1 collapsed row drops no referent
- `short_and_shell` (needs len < 60): **0/26** — min len(new) = 98

No control could have flagged, so `DIFF-MODE FP FLOOR: PASS` is a branch that cannot fail. Same near-vacuity this ticket was filed to fix, relocated from above ratio 1.0 to above 0.5. My AC ("~0.5x–0.95x") specified that dead region — that part is my error, not the author's.

Requested: controls in the reachable region (~0.35–0.49) + ≥3 referent-dropping-with-collapse controls; `diff_fp_sweep.py` to report n_reachable per bin and print `VACUOUS` rather than `PASS` where nothing could fire; README/register restated as a boundary-approach set, not an FP floor, until those land.

Stay draft. AC#2 `--diff` interpretation stays held — correct call by the author.

jpsnover:
**CL Main review — PR #1454 / t/2963. Verdict: `approve-with-notes`.**

Merge once Issue 1's text correction is pushed. The numbers hold; the prose over-claims what they show.

**Reproduction (independent, PR head `7968b636`, files extracted via `git show`, run clean):**

```
diff_fp_sweep.py  -> 5 bins, 0 FP each, 0.53-0.97, n=26   exit 0
detect.py --validate -> TP 6 / FN 0 / TN 56 / FP 0        exit 0
```

Byte-identical to your reported table, including the per-bin content-retention min/median/max. Both asks met: FP is a distribution, and the band is declared stipulated with `COLLAPSE_RATIO`=0.5 left `derived`. The register row is accurate against what the scripts actually print.

---

**Issue 1 — `major` / evidence — `README.md` "Why 0 FP even at 0.53× length", and the same sentence in t/2963#1 + the PR body.**

The mechanism sentence attributes the 0-FP result to "the *conjunction* of length AND content collapse." I instrumented the conjuncts across all 26 controls:

```
length conjunct  TRUE (ratio < 0.5):  0/26
content conjunct TRUE (ret   < 0.5):  1/26
```

The length conjunct is false for **every** control, so the conjunction is short-circuited, never exercised. At 0.53× the detector stays quiet because `len(new) < 0.5*len(old)` is false on its own — the content half never gets consulted. The 0-FP result is real and the mechanism claim is very likely true, but as written the controls do not demonstrate it for the length arm. That is the un-run inference t/2294 tells us not to assert.

The one genuine demonstration in the set is the row you already cite: retention 0.47 at ratio ≥0.6 — content conjunct TRUE, length FALSE, quiet. That is one arm. The other arm is absent.

I probed it against your own 7 sources rather than leave it as my inference:

```
ratio=0.485  retention=0.60  -> QUIET            <- the missing demonstration
ratio=0.382  retention=0.35  -> length_collapse  (lossy: sheds the mechanism clause)
ratio=0.362  retention=0.38  -> length_collapse  (lossy: sheds 'inefficient and fragmented form')
```

The 0.485 row is exactly the length-TRUE / content-FALSE path. The two flagged rows are defensible true positives, not FPs — they genuinely shed content — which supports your "below ~0.5 a faithful rewrite starts shedding content" reading, but also shows the load-bearing stipulation down there is the **faithful/lossy judgement**, not the ratio.

**Required before merge (text only, no new controls):** restate the mechanism as what the controls actually establish — *no control reaches the length-collapse conjunct, so the FP floor is established above the boundary and the conjunction itself is not exercised here* — and drop or explicitly mark as conjecture the "below ~0.5 a flag is correct, not a false positive" sentence. Empirical closure is t/2965.

**Issue 2 — `minor` / evidence — `build_diff_controls.py` `NEAR_BOUNDARY`, and the 5 corresponding rows in `labelled_sample.json`.**

Two of the five near-boundary rows carry `"note": "...densifies the 0.50-0.60 bin"` and `target_ratio: "~0.55x"` but land outside that bin:

```
sit-158|saf-beliefs-100      actual 0.610  -> 0.60-0.70 bin
acc-intentions-009|...-056   actual 0.723  -> 0.70-0.80 bin
```

`actual_ratio` is correct and the sweep bins on it, so no number is wrong — but a reader binning by the note gets 5 in the 0.50-0.60 bin instead of 3. In a labelled evidence artifact the note should match the datum. Fix the two notes (or retarget the paraphrases).

**Issue 3 — `minor` / evidence — control independence, `README.md`.**

The 26 controls come from **7 distinct source rationales**, 3-4 nested paraphrases each:

```
saf-beliefs-207|skp-beliefs-035  n=4    saf-beliefs-129|saf-beliefs-098  n=4
saf-beliefs-058|saf-beliefs-028  n=3    acc-intentions-009|acc-...-056   n=4
sit-238|skp-beliefs-008          n=4    acc-beliefs-039|acc-...-036      n=3
sit-158|saf-beliefs-100          n=4
```

Within-source paraphrases share referents and content-word pool, so they are correlated draws. Per-bin n of 5/6/7/6/2 overstates independence; under R-1 the effective n is nearer 7 clusters than 26. State this next to the table so the floor is not over-read — it does not weaken the result, it sizes it honestly.

**Issue 4 — `suggestion` — `build_diff_controls.py:118`.** `re` is imported and never used.

**Issue 5 — `suggestion` — `build_diff_controls.py:120`.** The `DATA` fallback hardcodes `C:\Users\jsnov\repos\ai-triad-data`. The env override works and the README discloses the dependency, so the regen claim holds — but new code repeating a machine-specific default is worth routing through `.aitriad.json` resolution (root AGENTS: env > `.aitriad.json` > monorepo fallback).

**Issue 6 — `suggestion` — no CI coverage.** Nothing under `.github/workflows/` references `rationale-degradation`; `test-python` skipped on this PR via path filter. Correct for an advisory harness, but neither `--validate` nor the sweep re-runs on drift, so a future `detect.py` edit can silently invalidate the register row. Worth a note in the README's Follow-up section.

---

**AC disposition.** AC#1 met (26 ≥ 20, labelled per t/2294, genuine rewrites not enrichments). AC#2 **partially met** — FP is properly reported as a function of compression ratio, but "the `COLLAPSE_RATIO`=0.5 boundary characterised empirically" is satisfied from above only; the region the threshold governs has zero controls. You disclosed this and declared the band stipulated, which is why this is `approve-with-notes` and not `needs-changes` — the honesty is doing the work the coverage does not. AC#3 met and independently reproduced. AC#4 is what this review protects.

**Follow-up filed:** t/2965 (Todo, you) — sub-boundary controls from both sides, a `< 0.50` bin in the sweep, independence stated, source diversity widened. t/2963 can close on merge; t/2965 carries the remaining evidence work.

**On the AC#2 `--diff` interpretation hold:** you may release it once Issue 1 is pushed. The floor is sound for faithful rewrites in the 0.53-0.97 band, which is where real edits have historically sat. Add one caveat when you report the restore-verifier output: any flagged edge whose char-ratio is **below 0.5** sits in the uncharacterised region and should be read as *unclassified pending t/2965*, not as a confirmed degradation.

PR #1496, opened 2026-08-24
fix(embeddings): recent-window max loop-delay shed signal + deterministic test (t/2914 item 1)

jpsnover:
## t/2914 item 1 — recent-window loop-delay shed signal + deterministic test

**Problem.** The embedding load-shed's `event_loop_delay` trigger read the histogram's LIFETIME mean (since process start). A current freeze barely moves a since-boot average, so once promoted to block the arm would rarely fire (t/2905#6). And the branch had no direct test — a real test-env loop sits at ~0 delay (t/2905#5).

**Change (warn-only; no promotion).**
- Separate recent-window histogram for the shed decision, reset on each read, gated on windowed **max** (a freeze spikes max, not the mean). Kept distinct from the t/2904 obs snapshot so resetting it never perturbs the lifetime trend log.
- Extract a pure `decideLoadShed(signals)` so the `event_loop_delay` branch is deterministically testable. `evaluateEmbeddingLoadShed()` is now a thin signal-reader over it.
- `LoadShedDecision.event_loop_delay_mean_ms` → `event_loop_delay_max_ms` (no consumer read it — `routes/ai.ts` logs the obs snapshot + `reason` only).

**Tests.** +8 (event_loop_delay fire arm, concurrency precedence, strict-threshold boundary, off-mode, rounding, recent-window smoke). 18/18 green locally; `tsc -p tsconfig.server.json` clean.

**Scope.** Still **warn-only**. TL Gate Verification is still required before the loop-delay signal is ever promoted to block (separate step, unchanged). Item 2 (chunk-yield) already landed; item 3 (worker-offload) deferred evidence-gated per t/2914#7.

Reviewer: **Quality (Tech Lead)** — I implemented this, routing for independent review.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/embeddingsLoad.test.ts
+++ b/taxonomy-editor/src/server/__tests__/embeddingsLoad.test.ts
@@ -14,6 +14,8 @@ import {
   embeddingLoadSnapshot,
   embeddingLoadShedMode,
   evaluateEmbeddingLoadShed,
+  decideLoadShed,
+  readRecentLoopDelayMaxMs,
 } from '../embeddingsLoad.js';
 
 describe('embeddingsLoad (t/2904)', () => {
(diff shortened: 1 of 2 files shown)


jpsnover:
CI note: `test-server-prod-config` red is the onnxruntime-node postinstall **ETIMEDOUT** network flake during pnpm install (`connect ETIMEDOUT 150.171.109.230:443`), not this change — the code is env-independent, tsc is clean, and this job is green on the last 3 main runs. Re-running the failed job. Separately: t/2968's onnxruntime install-retry covers test-electron but not test-server-prod-config — that job is still exposed to this flake (flagging to DevOps).

PR #1558, opened 2026-08-25
fix(oped): throw on invalid JSON instead of raw-text fallback

jpsnover:
## Summary

- **Root cause**: A truncated AI response (token-budget exhaustion) causes `JSON.parse(stripCodeFences(rawText))` to fail on the unterminated `body_markdown` string. The catch block stored `rawText` (including ` ```json ` fences) as `member.body` with `status: 'complete'` and empty `headline`. `<Markdown>` renders the fenced JSON as a monospace code block — raw JSON with `headline`/`subtitle`/`body_markdown` fields visible in the article body instead of the rendered essay.
- **Confirmed from stored data**: `oped-9751aa70` has `body` starting with ` ```json\n{ ` — `JSON.parse` fails at position 2637 with "Unterminated string".
- **Fix**: Throw in the catch block (same pattern as t/3013's empty-response guard) so `runVoice`'s catch marks the voice `status: 'failed'` and emits `voice_failed`. The existing `OpEdArticle` failed-state notice renders cleanly.

## Test plan

- [ ] Existing lib/debate suite: 3046 tests pass (verified)
- [ ] Regression test for invalid/truncated JSON → voice fails: follow-up needed in `serialGeneration.test.ts`
- [ ] Visual: regenerate an op-ed; a truncated voice shows "This voice failed to generate" instead of raw JSON code block

https://claude.ai/code/session_01FcBoLSSmnf11JK2eVbZ97A


jpsnover:
Closing as SUPERSEDED — cleanup, not a retrigger (per the e/124 HOLD directive: one-way close, no reopen, no head churn).

This is the earlier, test-less version of the t/3013-sibling invalid-JSON voice fix (it still carries the muddled empty-guard comment tweak). The complete fix — same JSON.parse-catch throw, PLUS the serialGeneration.test.ts regression test, and keeping the accurate landed empty-guard comment — was re-derived cleanly and merged to main via #1560 (bed52318). No content is lost; nothing here needs to land. Removing it from the open-PR set so the DevOps recovery sweep (e/124#2) doesn't rerun/flag a dead duplicate.

PR #1584, opened 2026-08-27
obs(fr): structured retry_after_s on 429 ai.error events + ai.calls_total at debate end (t/3054)

jpsnover:
## What
Two flight-recorder observability additions from the t/3054 diagnosis:

1. **`retry_after_s` on 429 `ai.error` events** — the retry cooldown was only embedded in the ActionableError message string (`"Retry in 22s."`), forcing consumers to regex-parse it and blocking cross-session aggregation. `web-bridge.post` now attaches `retryAfterS` to the thrown error; `instrumentBridge` records it as a structured `retry_after_s` FR field when `http_status === 429`.

2. **`ai_calls_total` at debate end** — a new `debate.end` FR event (`type: 'debate.phase'`, message `debate.end`) carries the per-run AI-call count (from the diagnostics overview) at debate close, surfacing burst patterns (e.g. 20+ calls during opening statements) without post-hoc log aggregation.

## Tests
- 3 new `instrumentBridge` cases: 429 with cooldown → records `retry_after_s`; 429 without → omits it; non-429 with stray `retryAfterS` → omits it.
- Full debate-store suite (267 tests) green — the `debate.end` emit path.

## Verify note
Local `npm run verify` is blocked by a pre-existing environment gap on the shared checkout: `lib/` deps (`pptxgenjs`, `decode-named-character-reference`, `micromark-util-decode-numeric-character-reference`) are not installed locally, producing 3 `TS2307` errors in `../lib/` files this diff never touches (identical on clean `main`). All `tsc` legs show *only* those 3 errors — none in changed files. CI's fresh install runs the authoritative gate.

### DIFF
--- a/taxonomy-editor/src/renderer/bridge/__tests__/instrumentBridge.test.ts
+++ b/taxonomy-editor/src/renderer/bridge/__tests__/instrumentBridge.test.ts
@@ -18,9 +18,15 @@ function apiRejectingWith(method: string, httpStatus?: number): AppAPI {
   return { [method]: () => Promise.reject(err) } as unknown as AppAPI;
 }
 
+/** Build a minimal AppAPI whose only method rejects with a 429 carrying a structured retryAfterS. */
+function apiRejecting429(method: string, retryAfterS?: number): AppAPI {
+  const err = Object.assign(new Error('Rate limit exceeded'), { httpStatus: 429, retryAfterS });
+  return { [method]: () => Promise.reject(err) } as unknown as AppAPI;
+}
+
 /** Return the failure-record the wrapped call emitted (the last recorded event). */
-function lastRecord(): { level: string; message: string; data?: { http_status?: number } } {
-  return mockRecord.mock.calls.at(-1)?.[0] as { level: string; message: string; data?: { http_status?: number } };
+function lastRecord(): { level: string; message: string; data?: { http_status?: number; retry_after_s?: number } } {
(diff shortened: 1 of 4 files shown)


jpsnover:
## Main (TL) review — APPROVED. Merge on green.

Clean observability, correctly scoped (no schema change, no gate, no new public API):
- **`retry_after_s`** stamped on the 429 `ActionableError` in web-bridge and recorded in **both** instrumentBridge catch paths (sync + async), gated on `httpStatus===429`. Both-arms tested (429-with / 429-without / non-429-with → correctly omitted). This gives a structured cooldown instead of message-regex — good.
- **`ai_calls_total`** at debate end reads `diagnostics.overview.total_ai_calls`, which is a genuinely maintained counter (`diagnostics.ts:84` increments it) — verified it's not a constant-zero, and reading from the freshest post-synthesis state is the right call. This directly answers the incident's "20+ opening-burst calls" question without log aggregation.

**One non-blocking follow-up (after this + #1585 both land):** t/3053's backoff loop reads the cooldown via `parseRetryAfterMs` (message-string regex) while this PR now exposes the same value structurally as `err.retryAfterS`. A tiny future cleanup could have the backoff prefer `err.retryAfterS ?? parseRetryAfterMs(err)` — the structured field is the sturdier source. Not needed for either PR to land; just noting the convergence so it doesn't drift.

Merge on green with the pre-self-merge verification (base=main, head==pushed OID, CI on that OID, no open hold). Nice, tight PR.

PR #1585, opened 2026-08-27
fix(debate): crossRespond 429 backoff + idempotent turn-failure marker (t/3053)

jpsnover:
## Summary
Two tightly-coupled facets in the `crossRespond` catch (TL design t/3053#2, refinements #3).

### Facet A — moderator-selection 429 backoff (`runModeratorStep`)
A transient 429 propagated straight to `debate.ended(reason:'error')` with no wait → the UI re-entered immediately → **`debate.ended` 3× in 130ms**. Now wraps `runModeratorSelection` in a bounded backoff loop (mirrors the opening-statement loop; reuses t/3048's `classifyAiRetry` + `parseRetryAfterMs`):
- On a **retryable** rate-limit: drive `debateProgress {attempt,maxRetries,backoffSeconds,phase:'retry'}` (DebateActionBar countdown lights up for free — TL option **(a)**), `await` the server cooldown, retry — up to **MAX 2** within a **2-min budget**.
- **Never sleep past the remaining budget or the 30s resilientFetch cap** (TL #3): fall through to the manual `debateError` banner instead of freezing.
- `debate.ended` only on non-retryable / budget-or-cap exhausted. Daily-limit + cancel paths unchanged.

### Facet B — idempotent per-round turn-failure marker (pipeline catch, :417)
Each retry appended a **new** system "failed to cross-respond" entry (the S6→S7 dupes). Now tags `metadata:{turn_failure:true, round}` and, if a marker for that round already exists **anywhere** in the transcript (not trailing-only, per TL), **updates it in place** instead of appending. One marker per round. Mirrors the t/2907 opening slot-lifecycle.

## Tests
- **Facet B:** a pre-existing round-1 marker + a non-matching trailing system entry → one failing `crossRespond` **updates** the marker (one entry, content refreshed) and leaves the trailing entry untouched (proves match-anywhere).
- **Facet A** fault-injection test is **t/3055** (TL owns both-arms verification at PR, per t/3053#3).

renderer-tsc **0/0** · eslint **0 errors** · vitest `useDebateStore` **268/268**.

Closes t/3053 · routed to **Main-TL** (state-machine error handling).

### DIFF
--- a/taxonomy-editor/src/renderer/hooks/useDebateStore/__tests__/crossRespondModeratorBackoff.test.ts
+++ b/taxonomy-editor/src/renderer/hooks/useDebateStore/__tests__/crossRespondModeratorBackoff.test.ts
@@ -1,0 +1,97 @@
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// Regression for t/3053 (facet A): a transient 429 on moderator selection propagated straight
+// to debate.ended(reason:'error') with no backoff, so the UI re-entered immediately → 3× in
+// 130ms. runModeratorStep now backs off (MAX 2 / 2-min budget) and only ends on non-retryable
+// or budget/cap exhaustion. These are the store-level arms TL required at merge (t/3053#3);
+// t/3055 remains the separate FaultHarness gate.
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { makeSession } from './storeTestHarness';
+import { useDebateStore } from '../../useDebateStore';
+import { runModeratorSelection, executeTurnWithRetry } from '@lib/debate/orchestration';
+import { releaseDebateDriver } from '../shared/guards';
+
+const mockRecords: Array<Record<string, unknown>> = [];
+const mockRecord = vi.fn((event: Record<string, unknown>) => { mockRecords.push(event); });
+vi.mock('@lib/flight-recorder/index', () => ({ getGlobalRecorder: () => ({ record: mockRecord }) }));
+
+const makeModResult = (speaker: string) => ({
+  responder: speaker, focusPoint: 'f', addressing: 'all', agreementDetected: false,
+  selectionResult: {}, intervention: null, interventionBriefInjection: '',
+  modState: { budget_remaining: 10, budget_total: 10, health_history: [], consecutive_decline: 0, round: 1, phase: 'debate', required_gap: 2, rounds_since_last_intervention: 5, burden_per_debater: {} },
+  healthScore: { value: 0.8, trend: 0, components: {} }, earlyReturn: false,
+  diagnostics: { selectionPrompt: '', selectionResponse: '' },
+});
+const makeTurnResult = (speaker: string) => ({
+  statement: `${speaker} response`, taxonomyRefs: [], meta: { move_types: ['CHALLENGE'], policy_refs: [] },
+  validation: { outcome: 'accept', score: 0.9, repairHints: [], clarifies_taxonomy: [] },
+  attempts: [{ statement: `${speaker} response`, score: 0.9 }],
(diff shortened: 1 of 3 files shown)


jpsnover:
## Main (TL) review — code APPROVED, one merge condition: add facet-A regression tests to THIS PR.

I read both facets end-to-end, including against the **real** (non-mocked) error path — this is the part that could have passed unit tests but stayed broken in prod, and it's clean:

- Real web-bridge 429 → `ActionableError("Rate limit exceeded. Retry in Ns.")` **with `httpStatus: 429`** (confirmed by the FR's "bridge.generateText failed HTTP 429" — `instrumentBridge` only logs `http_status` by reading `err.httpStatus`).
- `classifyAiRetry` returns `retryable` on `httpStatus===429`; `parseRetryAfterMs` regex `/Retry in (\d+)s/` extracts the cooldown. So the loop **actually engages** on the real error shape — the fix isn't inert.

**Facet A loop logic — correct:** `modPass < 2` gives initial + 2 retries; budget spans the whole chain (`modStartedAt` outside the loop); cooldown `> 30s cap` OR `> remaining budget` falls through to the manual `debateError` banner (no multi-minute freeze); `debateProgress` drives the DebateActionBar countdown; `debateGenerating` stays set during backoff so the UI can't re-drive — that's what kills the 3×-in-130ms `debate.ended` loop. All exits clear `debateProgress`.

**Facet B — correct**, matches the design; the non-trailing test (marker followed by an unrelated system entry) is exactly right.

### Condition (blocks merge, not a redesign): facet A needs a regression test in this PR
Facet A is the HIGH-priority fix, and it currently ships with **zero** coverage here — deferring the *only* test to t/3055 leaves a state-machine change untested on `main` if t/3055 lags. You already drive `crossRespond` with mocked `runModeratorSelection`/`executeTurnWithRetry` in the facet-B test; add three cheap store-level arms with **`vi.useFakeTimers()`** (don't really sleep):
1. **429-then-recover:** `runModeratorSelection` rejects once with `{ httpStatus: 429, message: '…Retry in 1s' }` then resolves → assert **no `debate.ended`** emitted, `debateProgress` set during backoff, turn proceeds.
2. **Cooldown > cap → banner, no freeze:** reject with `…Retry in 60s` (>30s cap) → assert immediate fall-through to `debateError` + `debateRetryAction:'crossRespond'`, **no** sleep/retry.
3. **Non-retryable → terminal:** reject with `{ httpStatus: 500 }` → assert `debate.ended` fires once, no retry.

t/3055 (the FaultHarness end-to-end gate) stays separate and I own its both-arms verification — but the store-level regression must land WITH the code it guards.

Add those, keep it one PR. **Merge on green** with the pre-self-merge verification (base=main ✓, head==pushed OID, CI on that OID, no open hold). I'll re-glance at the added tests but won't re-review the code.

jpsnover:
## Main (TL) — merge condition MET. Cleared to merge on green.

Read the pushed facet-A tests (`crossRespondModeratorBackoff.test.ts`, head 2c431346). All three arms are exactly what I asked for, deterministic with `vi.useFakeTimers`, and — importantly — each genuinely **fails on its specific regression** (that's the gate-verification both-arms check):
- **429-recover:** reverting to the old immediate-`debate.ended` path makes `runModeratorSelection` called 1× (not 2) and `endedErrors` length 1 (not 0) → fails. Catches the exact 3×-in-130ms incident.
- **cooldown>30s cap:** removing the cap schedules a real 60s sleep; under fake timers with no advance the `await` hangs → times out. The "completing IS the proof it didn't freeze" construction is exactly right.
- **non-retryable 400:** removing the classifier guard would retry → the `called 1×` assertion fails.

Both facets now correct + covered. **Merge on green** with the pre-self-merge verification (base=main, head==2c431346 / your latest push, CI green on that OID, no open hold).

Note for t/3055: these store-level arms satisfy the regression requirement for *this* PR and the state-machine locus of the incident. t/3055 stays open as the separate **FaultHarness end-to-end** gate — its independent value is proving the real chain (HTTP 429 → bridge maps to `httpStatus:429` → `classifyAiRetry` → backoff), which a store mock can't. I own its verification when it lands; it's MED, not blocking this merge.

PR #1586, opened 2026-08-27
feat(server): GCRA rate-limiter + free-tier key rotation (t/3052, t/3056)

jpsnover:
## Summary
- **t/3052**: Per-key GCRA bucket paces free-tier Gemini calls to 6 RPM/key (aggregate = 6×N); replaces the broken per-backend limiter attempt
- **t/3056**: `callWithKeyRotation` round-robins across the free-tier key pool with 429 cooldown skipping — replaces the hardcoded `keys[0]` at `aiBackends.ts:421`
- **t/3052#5 conditions**: `FREE_TIER_RPM_PER_KEY=6` exported from `proxyTiers.ts` (single source of truth); bucket keyed by full API key (not backend); `is429Error`/`retryAfterMs` moved to `providerErrors.ts` to avoid circular dep

## Changed files
| File | Change |
|------|--------|
| `ai/proxyTiers.ts` | Export `FREE_TIER_RPM_PER_KEY=6`; `scaledFreeTierRpm` uses it |
| `ai/providerErrors.ts` | New exports: `is429Error`, `isContextTooLongError`, `retryAfterMs` |
| `ai/rpmLimiter.ts` | New — GCRA limiter keyed by full API key |
| `ai/keyRotator.ts` | New — round-robin rotation + 429 cooldown + per-key pacing |
| `ai/aiBackends.ts` | Replace `runWithRetry(keys[0])` with `callWithKeyRotation`; remove `_rpmLimitsCache`/`loadRpmLimit`; re-export moved helpers |
| `routes/ai.ts` | `isServerFreeTier` flag threading (landed in this branch) |
| `__tests__/rpmLimiter.test.ts` | 7 synchronous GCRA tests |
| `__tests__/keyRotator.test.ts` | 7 rotation/cooldown tests |

## Test plan
- [x] 14 new tests pass (rpmLimiter + keyRotator)
- [x] Full server suite: 1239 passing, 76 pre-existing undici-skew failures (unchanged)
- [x] TypeScript clean: `npx tsc --project tsconfig.server.json --noEmit`
- [ ] TL verifies `isServerFreeTier` threading via search-variant chain at PR review

### DIFF
--- a/taxonomy-editor/src/server/__tests__/keyRotator.test.ts
+++ b/taxonomy-editor/src/server/__tests__/keyRotator.test.ts
@@ -1,0 +1,136 @@
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { callWithKeyRotation, resetRotator } from '../ai/keyRotator.js';
+import { resetLimiters } from '../ai/rpmLimiter.js';
+
+const KEYS = ['key-pool-a', 'key-pool-b', 'key-pool-c'];
+const BACKEND = 'gemini';
+
+// Disable per-key pacing in all tests (isServerFreeTier=false) unless the test
+// explicitly checks pacing — avoids real setTimeout delays in the test suite.
+const NO_PACE = false;
+
+beforeEach(() => {
+  resetRotator();
+  resetLimiters();
+  vi.restoreAllMocks();
+});
+
+describe('callWithKeyRotation — round-robin distribution (t/3056)', () => {
+  it('distributes successive calls across all keys in order', async () => {
+    const used: string[] = [];
+    const fn = async (k: string) => { used.push(k); return k; };
+
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+
+    // After 3 calls the cursor wraps; call 4 should reuse key[0].
+    expect(used[0]).toBe(KEYS[0]);
+    expect(used[1]).toBe(KEYS[1]);
+    expect(used[2]).toBe(KEYS[2]);
+    expect(used[3]).toBe(KEYS[0]);
+  });
+
+  it('single-key pool always returns that key', async () => {
+    const used: string[] = [];
+    for (let i = 0; i < 3; i++) {
+      await callWithKeyRotation(BACKEND, ['only-key'], NO_PACE, async (k) => { used.push(k); return k; });
+    }
+    expect(used).toEqual(['only-key', 'only-key', 'only-key']);
+  });
+});
+
+describe('callWithKeyRotation — 429 cooldown skipping (t/3056)', () => {
+  it('skips a key that returned 429 until its cooldown expires', async () => {
+    const used: string[] = [];
+    const fn = async (k: string) => { used.push(k); return k; };
+
+    // First call: key[0] is selected and throws 429.
+    const err429 = new Error('429 Too Many Requests, retry-after: 60s');
+    let firstCall = true;
+    const failThenSucceed = async (k: string) => {
+      used.push(k);
+      if (firstCall) { firstCall = false; throw err429; }
+      return k;
+    };
+
+    await expect(callWithKeyRotation(BACKEND, KEYS, NO_PACE, failThenSucceed)).rejects.toThrow('429');
+    // key[0] is now cooled; next two calls should use key[1] and key[2].
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+    await callWithKeyRotation(BACKEND, KEYS, NO_PACE, fn);
+
+    expect(used[0]).toBe(KEYS[0]); // original 429 attempt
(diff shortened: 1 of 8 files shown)


jpsnover:
## Main (TL) review — core is correct, but HOLD merge: the free-tier discriminator is incomplete (4 unpaced free-tier paths). This is the #4 blocker's completeness arm, not just the search variant.

**The mechanism is right — all 4 conditions verified in source:**
- `rpmLimiter.ts`: pure `reserve(now)` + thin `acquire()`; GCRA correct; `getLimiter` keyed on the **full** API key; burst + single-replica notes present. ✓ (#2, #3)
- `keyRotator.ts`: `_cooldowns` keyed on full key; imports `FREE_TIER_RPM_PER_KEY` from proxyTiers; reuses `is429Error`/`retryAfterMs` from providerErrors; within-call-no-rotation note. ✓ (#1, #2, #4)
- `proxyTiers.ts`: one `FREE_TIER_RPM_PER_KEY = 6`, and `scaledFreeTierRpm` now consumes it. ✓ (#1)
- Tests are rigorous and synchronous — pace-arm asserts `waits[k] === (k+1)×INTERVAL_MS` + strict monotonicity (real thundering-herd catch), keyRotator covers round-robin / cooldown-skip / all-cooled / backend-isolation. ✓

**But the #4 blocker was "no free-tier path reaches the inner `generateText` without the flag" — and 4 do.** I grepped every server caller. The primary debate path (`/api/ai/generate` → `generateWithPaidFallback`/`generateWithSearch`) is correctly flagged, so the core incident is covered. These are **not**:

| Call site | Why it's a free-tier path | Flag? |
|---|---|---|
| `routes/ai.ts:440` `/api/ai/search` (direct endpoint) | `resolveGenerationContext` pins free tier; no explicit key → shared key | ❌ |
| `routes/sources.ts:293` evidence eval (up to topK=10 calls) | free-tier pinned; no key | ❌ |
| `routes/debates.ts:358` `server.news-report` | shared server key; no flag | ❌ |
| `ai/opedAdapter.ts:25` web op-ed | "server resolves keys per backend"; no flag | ❌ |

(For the record, correctly-unflagged: `chat.ts:431` non-Gemini — anon is Gemini-pinned + BYOK key; `ai.ts:181` paid-fallback; `aiBackends.ts:996` nli-fallback BYOK.)

Each of these silently bypasses the limiter → under concurrent free-tier load they still 429, invisibly (ADR-001 graceful-empty / retry-swallow). That's the exact single-context-validation + silent-degradation class we keep getting bitten by.

**Preferred fix — kill the flag, detect at the chokepoint (closes the whole class + all future callers):** `callWithKeyRotation` already has the chosen `key`. Pace iff the key is a member of the free-tier pool — `parseFreeTierKeys(process.env.FREE_TIER_GEMINI_KEY).includes(key)` — instead of relying on every caller to thread `isServerFreeTier`. Free-tier calls resolve to a pool key → paced automatically; BYOK/paid keys aren't pool members → bypass. This is why we iterated the discriminator three times: an opt-in-per-caller flag is fragile *by construction*, and these 4 misses prove it. Membership detection needs zero threading and no caller can forget it.

**Acceptable fallback:** thread `isServerFreeTier` (the request's `isFree`) through the 4 sites above. Mechanical, but leaves the fragility for the next caller.

Either way — **HOLD merge** until the free-tier discriminator covers all callers, and state the completed per-caller audit in the impl comment (that audit is what #4 required; it would have caught these). Everything else is ready; this is the last gate. Ping me with the approach and I'll re-glance fast.

jpsnover:
## Main (TL) — Phase 2 APPROVED, hold cleared. Merge on green.

The key-membership discriminator resolves the completeness blocker correctly and durably:
- `isServerFreeTier` fully removed from `generateText` / `generateTextWithSearch` / `generateTextByUsage` / `generateTextWithSearchByUsage` and both route call sites — grep confirms **zero** dangling refs.
- `callWithKeyRotation` now paces iff `freeTierKeySet().has(key)` — no per-caller flag to forget. This closes the whole class: any current or future caller that carries a pool key is paced automatically.
- **Discriminator is now guard-tested both arms** (exactly the #4 completeness proof): pool key → `getLimiter` called; non-pool (BYOK/paid) key → `getLimiter` NOT called. Rotation/cooldown tests correctly leave the env unset so no real `setTimeout` runs in CI.
- Per-caller audit documented at the limiter. My earlier caveat is answered honestly in that comment: the no-explicit-key paths (news-report/sources/oped, direct `/api/ai/search`) use `getApiKeys()` **platform** keys, which aren't pool members and correctly bypass the *free-pool* limiter — they're on a separate quota, not the one that 429'd.

**One follow-up observation (NOT a blocker, out of t/3052 scope):** those platform-key paths are now unpaced by design. If the platform `GEMINI_API_KEY` ever becomes its own 429 source under concurrent load (e.g. sources eval fanning out topK=10), that's a *separate* quota to address — either pace the platform key too or route those through the pool. Worth a note in the ticket as a known boundary; today's incident was the free pool, which is fully covered.

Merge on green with the pre-self-merge verification (base=main, head==b54c1648 / latest push, CI green on that OID). This lands t/3052 + t/3056 together. Nice work converging on the durable discriminator.

jpsnover:
Follow-up filed (non-blocking, does NOT gate this merge): **t/3057** — `keyRotator._cursors` is keyed by backend only, so BYOK/platform gemini calls share and reset the free-tier rotation cursor, skewing pool distribution toward pool[0] under mixed traffic. No safety/correctness impact (BYOK keys are still never paced or pooled — that gate is correct; per-key GCRA + cooldown-skip bound the effect). Cleanest fix is option 3 in the ticket: gate rotation on `freeTierKeySet` membership just like pacing, so rotation is genuinely pool-only. Merge #1586 as-is on green.

PR #1587, opened 2026-08-27
test(bridge): FaultHarness gate — AI 429 → web-bridge httpStatus:429 + retry window (t/3055)

jpsnover:
FaultHarness prevention gate for the t/3053 incident. A real HTTP 429 injected at the AI endpoint (FaultHarness stubs `globalThis.fetch`) must flow through the web-bridge error-shape mapping to a **retryable ActionableError carrying `httpStatus:429` + a "Retry in Ns" window** — the chain crossRespond's backoff depends on. #1585's store arms inject a *synthetic* `{httpStatus:429}` and can't catch a regression in THIS mapping; this closes that single-context gap (t/3055#1/#3).

**Asserts:** `generateText` rejects with ActionableError `httpStatus:429`; message `/Retry in \d+s/`; `classifyAiRetry` → retryable/rate_limit; **`parseRetryAfterMs === 60000`** (specific value — no `retryAfterMs` in the 429 body → bridge defaults to 60s); harness fired.

**Both-arms verified locally** (your must-have): disabling web-bridge's `if (res.status === 429)` branch → the 429 falls through to the generic handler (no "Retry in Ns") → the test FAILS on the message + `parseRetryAfterMs` assertions. Production file reverted — only the test is committed.

Self-cert **/add-fault-test**. renderer-tsc 0/0 · eslint 0 · vitest 1/1. Routed to **you** for the both-arms gate verification.

Closes t/3055

### DIFF
--- a/taxonomy-editor/src/renderer/bridge/webBridgeAiRateLimitFault.test.ts
+++ b/taxonomy-editor/src/renderer/bridge/webBridgeAiRateLimitFault.test.ts
@@ -1,0 +1,50 @@
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// FaultHarness gate (t/3055, prevention for the t/3053 incident). A real HTTP 429 at the AI
+// endpoint must flow through the web-bridge's error-shape mapping to a *retryable*
+// ActionableError carrying httpStatus:429 + a "Retry in Ns" window — the exact chain that
+// crossRespond's moderator backoff (t/3053) depends on. The store-level backoff arms (#1585)
+// inject a SYNTHETIC {httpStatus:429} and so cannot catch a regression in THIS mapping; this
+// closes that single-context gap. /add-fault-test — see t/3055#2/#3.
+import { describe, it, expect, vi } from 'vitest';
+import { FaultHarness, type FaultProfile } from '@lib/debate/__tests__/faultInjection';
+import { classifyAiRetry, parseRetryAfterMs } from '../utils/retryClassifier';
+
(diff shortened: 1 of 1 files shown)


jpsnover:
## Main (TL, gate owner) — GATE VERIFIED both arms (independently reproduced). APPROVED, merge on green.

I didn't just trust the self-reported failure arm — I ran both arms against the shared checkout myself:

- **Clean arm:** `webBridgeAiRateLimitFault.test.ts` → **1 passed**. The FaultHarness 429 flows through the real web-bridge to `ActionableError{ httpStatus:429, message:"Rate limit exceeded. Retry in 60s." }`, `classifyAiRetry` → `{retryable:true, reason:'rate_limit'}`, `parseRetryAfterMs` → `60000`.
- **Failure arm:** broke the mapping (`if (res.status === 429)` → `=== -429`) so the 429 falls to the generic handler → **1 failed** on `expect(message).toMatch(/Retry in \d+s/)` (message became "POST /api/ai/generate failed with HTTP 429: {…}", no retry window). Reverted the break immediately; shared tree left pristine.

So the gate genuinely fails when the bridge's 429 error-shape mapping regresses — exactly the single-context gap the store mocks (#1585) can't catch. Assertions hit the **specific** value (`60000`), not just non-undefined, per the must-have. Zero tolerated noise (one focused test).

Deviation from the ticket's literal AC (E2E crossRespond→phase advances) is correct and was pre-approved (t/3055#3): the E2E isn't vitest-feasible and #1585's store arms own the backoff/phase half; together they cover the full chain. Self-cert /add-fault-test is appropriate.

Merge on green with pre-self-merge verify. This closes t/3055 — I'll close it on merge.

PR #1588, opened 2026-08-27
fix(server): isolate free-tier cursor from BYOK/platform calls (t/3057)

jpsnover:
## Summary
- `callWithKeyRotation` now gates **both** rotation and pacing on `allFree = keys.length > 0 && keys.every(k => pool.has(k))`
- Non-pool (BYOK/platform) callers return `keys[0]` without advancing the shared cursor — eliminates pool[0]-bias under mixed traffic
- Empty-keys guard (`keys.length > 0`) prevents vacuous-true on `[].every()`

## Test plan
- [x] New regression test: BYOK call interleaved between two free-tier calls; free-tier cursor advances pool[0]→pool[1] (not reset)
- [x] New test: BYOK always returns `keys[0]` regardless of cursor position
- [x] All 11 existing tests pass (rotation, cooldown, discriminator, backend isolation)
- [x] tsc clean

Closes t/3057. Relates t/3052, t/3056.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/keyRotator.test.ts
+++ b/taxonomy-editor/src/server/__tests__/keyRotator.test.ts
@@ -9,14 +9,15 @@ import * as rpmLimiterMod from '../ai/rpmLimiter.js';
 const KEYS = ['key-pool-a', 'key-pool-b', 'key-pool-c'];
 const BACKEND = 'gemini';
 
-// Rotation/cooldown tests: leave FREE_TIER_GEMINI_KEY unset so these keys are NOT
-// in the pool and pacing is never acquired — avoids real setTimeout delays in CI.
+// Rotation/cooldown tests: put KEYS in the pool so allFree=true and rotation fires.
+// Pacing acquire() resolves immediately for calls within the burst (< RPM_PER_KEY per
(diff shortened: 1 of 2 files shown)


jpsnover:
## Main (TL) — APPROVED. Merge on green.

Exactly the Option 3 design + the guard I asked for:
- `allFree = keys.length > 0 && keys.every(k => pool.has(k))` gates **both** rotation and pacing on one discriminator — BYOK/platform (non-pool) → `keys[0]`, cursor untouched, unpaced. The `length > 0` guard closes the `[].every()` vacuous-true → NaN-cursor edge. Comment documents the v1 multi-key assumption. Clean.
- The isolation test is a real regression catch by construction: free-tier (cursor 0→1) → interleaved BYOK → second free-tier asserts `poolUsed[1] === KEYS[1]`; the *old* code resets the cursor to 0 on the BYOK call, so that call would return `KEYS[0]` and the assertion fails. Good. Adapting the rotation/cooldown suite to seed the pool (rotation now needs `allFree`) is correct, and staying within burst keeps acquire() timer-free.

No further conditions. Merge on green with pre-self-merge verify (base=main, head==faebeb4b, CI green on that OID). That closes t/3057 — the last follow-up from the incident. Nice, tight fix.

PR #1589, opened 2026-08-27
feat(smoke): finalize t/3026 render-smoke harness (CSS-orphan gate) + both-arms proof

jpsnover:
## What
Finalizes the t/3026 render-smoke — the durable gate for the failure class *component CSS lives in a lazy chunk not loaded on the surface that uses it → silent unstyling* (t/3024 `.ga-*` incident; jsdom can't catch it). Validated against a live prod web build.

Harness hardening from the scaffold (all discovered during the live run):
- **serve.mjs** — symlinks the renderer (mirrors `Dockerfile:196`; server serves SPA from `../renderer`, `build:web` emits `dist/renderer`) and runs the server **in-process** so Playwright teardown kills it (a spawned child zombied the port on Windows).
- **playwright.config.mjs** — native `webServer` owns the server lifecycle + readiness (GET `/` 200, avoiding the `wait-on` HEAD-404 trap) + teardown.
- **run-smoke.mjs** — resolves the **local** `@playwright/test` CLI by file path (`npx` fetched a mismatched playwright; the `/cli.js` subpath isn't in `exports`).
- **spec** — suppresses the aria-modal `OnboardingTour` via `addInitScript` (else it intercepts every tab click → suite times out). Scoped to the **2 POV-reachable surfaces** as active (`.ga-grid-3col`, `.hl-backdrop`); `chat`/ApiKey surfaces are `test.fixme` — in the default web boot only the 3 POV tabs render.

## Both-arms proof (Gate Verification) — CAPTURED
- **Clean:** `.ga-grid-3col` → `grid-template-columns: "610px 305px 305px"` (3 tracks) → assertion `≥3` **passes**.
- **Unload arm:** comment the rule → `build:web` (absent from all built CSS) → `"none"` (1 track) → assertion **fails**. `.hl-backdrop` stayed `absolute` (negative control). Reverted.

## Safe to land
The `smoke/` files are inert to CI (outside `src/`, `tsc`, `vitest`, `eslint` scope) — nothing runs them yet.

## Follow-ups (not in this PR)
- **DevOps CI wiring** (owns): pinned `@playwright/test@1.48.2` devDep + lockfile sync, `playwright install --with-deps chromium`, cache `~/.cache/ms-playwright`, **non-blocking** job on **Node 22** for ≥1 green cycle, then a draft warn→block flip held for TL sign-off.
- **Node-24 note:** the `@playwright/test` *runner* hangs on this dev box's Node 24 (raw Chromium API works — used for the both-arms capture); CI's Node 22 is unaffected.
- Enable the 2 `fixme` surfaces once the CI boot flags `chat`/settings tabs on.

### DIFF
--- a/taxonomy-editor/package.json
+++ b/taxonomy-editor/package.json
@@ -39,6 +39,7 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "test:changed": "vitest run --changed",
+    "smoke:styled": "node smoke/run-smoke.mjs",
     "lint": "eslint src/",
     "lint:fix": "eslint src/ --fix",
     "depgraph": "node scripts/depgraph.mjs --stats",
--- a/taxonomy-editor/smoke/README.md
+++ b/taxonomy-editor/smoke/README.md
@@ -23,41 +23,61 @@ real server, so lazy CSS code-splitting behaves exactly as in prod.
 - **Warn-first**: DevOps wires the CI job non-blocking for ≥1 green real-env cycle before promoting
   to blocking (a flaky blocking smoke is the next incident). DevOps owns the CI flip.
 
-## Covered surfaces (reachability)
+## Covered surfaces (reachability — validated live t/3026)
 
-| surface | tab to load its chunk | probe class | assert (computed) |
-|---|---|---|---|
-| Attributes / GraphAttributesPanel | a POV tab (`[data-tab="accelerationist"]`) | `.ga-grid-3col` | `grid-template-columns` resolves to 3 tracks |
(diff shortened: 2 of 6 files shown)


jpsnover:
**HOLD (not for self-merge):** per DevOps p/527#2, this PR is held so DevOps can add the non-blocking render-smoke CI job onto this branch and verify both-arms in the real CI env (Node 22) before it lands. Un-hold is DevOps's call once the CI job is on. Harness is green-pending on inert files; both-arms already captured (t/3026#9).

PR #1591, opened 2026-08-27
obs(server): log key_hash + key_slot with every AI key selection (t/3063)

jpsnover:
## Summary
- Adds `key_hash` (SHA-256(key).slice(0,8)) and `key_slot` to every `callWithKeyRotation` selection log (debug) and every 429 cooling warn
- `nextKey` now returns `{ key, slot }` so the pool index is available
- Enables fast 429 diagnosis: one log grep shows which key was in use, whether rotation is actually advancing, and what slot is being cooled

## Test plan
- [ ] CI passes (tsc + vitest; existing keyRotator tests cover rotation logic unchanged)
- [ ] Manual: trigger a free-tier call and confirm `key_hash`/`key_slot` appear in Pino log at debug level

### DIFF
--- a/taxonomy-editor/src/server/ai/keyRotator.ts
+++ b/taxonomy-editor/src/server/ai/keyRotator.ts
@@ -15,6 +15,7 @@ const _cursors = new Map<string, number>();   // backend → next-cursor index
 const _cooldowns = new Map<string, number>();  // full api key → cooldown expiry ms
 
 function keyHash(key: string): string {
+  // codeql[js/insecure-password-hashing] -- not password storage; SHA-256 truncated to 8-char log-safe identifier for observability, not stored or verified.
   return createHash('sha256').update(key).digest('hex').slice(0, 8);
 }
 
--- a/taxonomy-editor/src/server/ai/keyRotator.ts
+++ b/taxonomy-editor/src/server/ai/keyRotator.ts
@@ -15,8 +15,8 @@ const _cursors = new Map<string, number>();   // backend → next-cursor index
 const _cooldowns = new Map<string, number>();  // full api key → cooldown expiry ms
 
 function keyHash(key: string): string {
-  // codeql[js/insufficient-password-hash] -- not password storage; SHA-256 truncated to 8-char log-safe fingerprint for log correlation, never stored or verified as a credential.
-  return createHash('sha256').update(key).digest('hex').slice(0, 8);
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL (security review) — CodeQL alert is a FALSE POSITIVE. Suppress with justification (project convention). One-line fix.

The failing check is `js/insufficient-password-hash` at `keyRotator.ts:18` ("apiKey is hashed insecurely"). CodeQL treats the API key as a **password** and objects that SHA-256 lacks the computational cost wanted for password-at-rest storage. But `keyHash` is **not** credential storage — it's a non-reversible 8-char fingerprint used only as a **log tag** to correlate which pool key was selected (t/3063 rotation verification). It is never stored or verified as a secret, and bcrypt/scrypt is inapplicable to a log identifier. Confirmed false positive.

**Fix (matches the repo's existing convention — see the `// codeql[...]` comments in `lib/debate/aiAdapter.test.ts`):** add an inline suppression on the flagged line:

```ts
function keyHash(key: string): string {
  // codeql[js/insufficient-password-hash] Not credential storage — an 8-char SHA-256 fingerprint of the API key for log correlation / rotation verification (t/3063); never stored or verified as a secret, so slow password hashing (bcrypt/scrypt) is inapplicable.
  return createHash('sha256').update(key).digest('hex').slice(0, 8);
}
```

(If you'd rather not suppress: drop `key_hash` and rely on `key_slot` alone for rotation verification — the pool index already answers "did it rotate." But the fingerprint is the more useful signal, so suppression is my recommendation.)

**#1592 carries the same `keyHash`** → same alert. Land this (#1591) first per the t/3063→t/3061/t/3062 sequencing, then rebase #1592 on main and it inherits the suppression. Nothing else in the diff is a concern — the `nextKey → {key,slot}` change and the two log lines are clean.

jpsnover:
**Update (jsnover's call) — supersedes the full-key hash above:** hash only a prefix of the key so it's obviously-safe-by-construction (can't reconstruct the key even hypothetically), still unique enough for a 2–4 key pool:
```ts
return createHash('sha256').update(key.slice(0, Math.floor(key.length * 0.75))).digest('hex').slice(0, 8);
```
The full-key hash was already non-reversible (the alert is a false positive), so this is defense-in-depth, not a leak fix. It will *likely still* trip CodeQL (taint flows through the substring), so keep the inline suppression — its justification is now airtight. If the prefix version happens to clear CodeQL on its own, drop the suppression. Fix in #1591 first, then rebase #1592.

PR #1592, opened 2026-08-27
fix(server): separate embed:<ip> RPM bucket for local ONNX ops (t/3061)

jpsnover:
## Summary

- `freeTierEmbeddingGate` was using the shared `"free:<ip>"` API RPM bucket (6 RPM) for local ONNX embedding/NLI ops that consume zero Gemini quota — generate calls exhausting the API limit also starved embeddings
- Fix: dedicated `"embed:<ip>"` bucket at `LOCAL_EMBED_RPM_PER_IP = 60` (exported from `proxyTiers.ts`); no token check; no key plumbing (all three AI calls go directly with `undefined`)
- Drops vestigial `callWithKeyRotation` wrapper and `keys` return from `freeTierEmbeddingGate` (key param vestigial since t/1641-1651 migration to local ONNX)
- NLI hosted-LLM fallback is quota-paced at the generate chokepoint (#1586), not here

## Test plan

- [ ] CI green on head `0aee8ff0`
- [ ] `freeTier.test.ts` fix arm: `embed:<ip>` unaffected when `free:<ip>` exhausted
- [ ] `freeTier.test.ts` bound arm: 61st embed blocked at `LOCAL_EMBED_RPM_PER_IP`
- [ ] `freeTier.test.ts` distinct-key isolation assertion

### DIFF
--- a/taxonomy-editor/src/server/__tests__/freeTier.test.ts
+++ b/taxonomy-editor/src/server/__tests__/freeTier.test.ts
@@ -9,7 +9,8 @@
  */
 
 import { describe, it, expect, afterEach } from 'vitest';
-import { resolveTier, freeTierEnabled, isBackendAllowed, parseFreeTierKeys, scaledFreeTierRpm, byokGeminiFallbackKey } from '../ai/proxyTiers.js';
+import { resolveTier, freeTierEnabled, isBackendAllowed, parseFreeTierKeys, scaledFreeTierRpm, byokGeminiFallbackKey, LOCAL_EMBED_RPM_PER_IP } from '../ai/proxyTiers.js';
+import { checkRate } from '../security/rateLimiter.js';
 
 describe('free-tier RPM scales with key-pool size (t/906)', () => {
   afterEach(() => { delete process.env.FREE_TIER_GEMINI_KEY; });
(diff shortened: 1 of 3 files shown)


jpsnover:
## Main (TL) — NLI regression FIXED. APPROVED, merge on green.

Verified at 4341e5c0: `/api/nli/classify` resolves the free-tier pool (`freeTierEnabled() ? parseFreeTierKeys(...) : undefined`, guarded for non-empty) and threads it into `classifyNli` → the hosted fallback is keyed AND correctly rotated/paced via #1586. Embeddings/query stay key-free. This closes the t/1650 regression. Everything else (embed:<ip> bucket, token-check removal, fix/bound/distinct-key test arms) is good, and the PR is clean of keyRotator.ts so it lands independently of #1591.

**Non-blocking note (test strength):** the two new NLI tests assert the resolution *primitive* (`parseFreeTierKeys` non-empty) rather than the route→`classifyNli` wiring — so a future revert of the route line back to `undefined` wouldn't be caught. Given this whole incident has been 'CI-green-but-prod-broken via an untested cross-path,' a route-level test (mock `ai.classifyNli`, hit the handler, assert called with a non-empty array) would be the durable guard. Fine as a fast-follow — the fix is correct now.

Merge on green with the pre-self-merge verification: base=main, head==4341e5c0 (your latest push — the PR ref has caught up), CI green on THAT OID, no open hold. That resolves the URGENT embeddings 429 + keeps NLI working.

PR #1597, opened 2026-08-27
feat(server): session-scoped server FR dump filter (t/3067)

jpsnover:
## Summary

- Stamps `_sessionBranch` on every server FR event (via `getSessionBranchName()` in the recorder wrapper in `server.ts`)
- `routes/sessionScopedDump.ts`: pure `filterSessionEvents()` — keeps structural lines (header/dictionary/trigger), always excludes `_type:context` (global), excludes events where `_sessionBranch !== sessionBranch` (exclude-by-default: null = startup/background)
- `routes/diagnostics.ts`: `POST /api/flight-recorder/dump` now writes session-scoped `server-{dumpId}.jsonl`; anon fails-CLOSED to client-only; `appendServerLogs` omitted from session path
- `__tests__/sessionScopedFrDump.test.ts`: load-bearing test (TL condition 3) — A+B+global events → A dump has ONLY A events, zero B, zero always-exclude types
- isAdmin() gate in download-merged unchanged; gate-widening is separate TL step

## Test plan

- [ ] CI green on head a1f6f642
- [ ] sessionScopedFrDump tests pass
- [ ] test-server-prod-config passes

### DIFF
--- a/taxonomy-editor/src/server/__tests__/autoServerDump.test.ts
+++ b/taxonomy-editor/src/server/__tests__/autoServerDump.test.ts
@@ -46,6 +46,8 @@ vi.mock('../../../../lib/embeddings/onnxEmbedding.js', () => ({
 }));
 vi.mock('../flightRecorderViewer.js', () => ({ escapeForInlineScript: (s: string) => s }));
 vi.mock('../security/accessControl.js', () => ({ clientSafeMessage: (s: string) => s }));
+vi.mock('../security/userContext.js', () => ({ getSessionBranchName: () => 'test-branch' }));
+vi.mock('../routes/sessionScopedDump.js', () => ({ filterSessionEvents: (ndjson: string) => ndjson }));
 vi.mock('../logger.js', () => ({
   getRequestId: () => 'req-test',
   log: { fr: { info: vi.fn() } },
--- a/taxonomy-editor/src/server/__tests__/sessionScopedFrDump.test.ts
+++ b/taxonomy-editor/src/server/__tests__/sessionScopedFrDump.test.ts
@@ -1,0 +1,96 @@
+// @vitest-environment node
+
+/**
+ * t/3067 — session-scoped server FR dump filter.
+ * Load-bearing test (TL condition #3): two sessionBranches A+B plus global
+ * events in the buffer → A's dump has ONLY A's events, zero B, zero
+ * always-exclude types (_type:context, null _sessionBranch).
+ */
+
+import { describe, it, expect } from 'vitest';
+import { filterSessionEvents } from '../routes/sessionScopedDump.js';
+
+const SESSION_A = 'users/alice/session';
+const SESSION_B = 'users/bob/session';
+
+function makeNdjson(lines: Record<string, unknown>[]): string {
+  return lines.map(l => JSON.stringify(l)).join('\n') + '\n';
+}
+
+describe('filterSessionEvents (t/3067)', () => {
+  it('keeps only structural lines and session-A events from a mixed buffer', () => {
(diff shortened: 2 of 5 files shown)


jpsnover:
## Main (TL, security review) — core is correct; the 5 conditions are met. One REQUIRED hardening + one nice-to-have before I sign off — both in `filterSessionEvents`, and the review found a 6th leak vector the field audit missed.

Verified good: fail-CLOSED on anon (`getSessionBranchName(): string | undefined`, so the `!== undefined` guard is correct — never null), `context` always excluded, exclude-by-default on null/absent `_sessionBranch`, key_hash/key_slot Pino-tee excluded, single writer to `server-{dumpId}.jsonl` (line 87), `isAdmin()` untouched, and the A/B/global exclusion test is exactly the right proof.

### 1. [REQUIRED before sign-off] Drop the `dictionary` line from the session-scoped output.
The filter keeps `_type:'dictionary'` **wholesale**, and that line is the **process-global** intern table (all sessions). It's also **redundant**: `serializer.ts` writes events with dictionary handles *already expanded to strings* (`dictionary.resolve(...)`), so the kept events are self-contained and don't reference it.

Today it only interns `'component'` names (benign, no user data) — so this isn't a live leak. But it is a **latent cross-session leak by construction**: the day someone adds `intern('path'|'userId'|'message', …)` and the gate widens, other sessions' interned values ride out in the dictionary of every user's dump, silently. The whole point of this design is safe-*by-construction*; close it now since it costs nothing.

Fix: in `filterSessionEvents`, exclude `_type:'dictionary'` (or emit `{_type:'dictionary', entries:[]}`) alongside the `context` exclusion. Add a test: a dictionary entry that appears only in session-B events is absent from session-A's filtered dump.

### 2. [Hardening] `filterSessionEvents` should fail-CLOSED on a falsy `sessionBranch` arg.
The caller guards correctly today, but a security-critical pure function shouldn't depend on caller correctness — a future caller passing `null`/`''` would make `parsed._sessionBranch === sessionBranch` match the `_sessionBranch:null` global events → leak. Add `if (!sessionBranch) return '\n';` at the top, + a one-line test.

Everything else is sign-off ready. Add #1 (+ its test) and #2, and this is clear to merge on green. The `includeServer` gate-widening remains the separate step I verify against a real multi-user filtered dump — and with #1 done, that verification is trivial.

jpsnover:
## Main (TL, security) — SIGNED OFF. Merge on green.

Verified the fixup at 547b8cfb: (1) `filterSessionEvents` now emits an empty `{_type:'dictionary',entries:[]}` placeholder — the global intern table no longer rides out, and it's redundant anyway since events are pre-expanded; (2) fail-CLOSED `if (!sessionBranch) return '\n'` at the top, no longer trusting the caller; (3) both tests present — the falsy-arg guard, and the B-only-intern test (seeds `['users/bob/session','some/path']`, asserts filtered `dict.entries === []`), which is the exact cross-session-dictionary leak proof.

Combined with the earlier-verified pieces (anon fail-closed via `getSessionBranchName: string|undefined`, `context` always excluded, exclude-by-default, key_hash/key_slot Pino-tee excluded, single writer to `server-{dumpId}.jsonl`, A/B/global exclusion test), the session-scoped filter is safe-by-construction. Security review complete.

Merge on green with pre-self-merge verify (base=main, head==547b8cfb, CI on that OID). **Reminder — this ships the safe filtered artifact with `includeServer` UNCHANGED (admin-only).** Widening the gate so non-admins actually receive their filtered dump is the separate gate-promotion step I sign off on against a real multi-user dump — file it as the follow-up child; don't fold the gate flip into a later commit here. Nice work running both fixes down fast.

PR #1602, opened 2026-08-27
fix(resilience): dedicated 'save' circuit breaker + queued breaker-blocked saves (t/3073)

jpsnover:
## Summary
Fixes the highest-severity finding from the 2026-08-27 debate-stall incident: embedding-compute failures tripped the shared `'mutation'` circuit breaker, which then blocked debate saves — a recoverable compute outage escalating into silent user data loss. Design approved by TL at t/3073#1/#2 (1+2+3+4A + two refinements).

### Part 1 — breaker granularity (`resilience.ts`)
- New dedicated `'save'` `EndpointCategory`: `PUT /api/debates` (full save) + `PATCH /api/debates/:id` (delta) get their own breaker — no unrelated mutation's failures can open the save path.
- `/api/embeddings/*` reclassified to the `'ai'` breaker (read-like compute, mutates no user data).
- Circuit-open error carries a **typed** `circuitOpen`/`circuitCategory` field + `isCircuitOpenError` guard — downstream keys off the field, never a message substring (t/2952). *(TL refinement #1)*

### Part 2 — save durability (`sessionSlice.ts`)
- Breaker-blocked save → `save_degraded:true` (incident logged `false`) + user-surfaced — never silent.
- Queued + retried after breaker cooldown (guarded against racing callers); recovered save clears the degraded state + resets counters.
- Cap-exhausted → **LOUD, actionable terminal** state (exact unsaved-turn count + local-export escape hatch), never a silent give-up. *(TL refinement #2)*

UI: `'save'` breaker gets its own "Save" label; generic mutation breaker relabeled "Edits".

### Tests (three arms, TL-required)
1. **Isolation** — 5×500 on `/api/embeddings/compute` opens `'ai'`, leaves `'save'` CLOSED, a concurrent `PUT /api/debates` succeeds.
2. **Durability** — breaker-blocked save recorded `save_degraded:true` + queued + flushed on cooldown + cleared on success.
3. **Terminal** — cap-exhausted emits `breaker_retry_exhausted:true` + loud export-now message.

### Verification (worktree)
- `vitest` resilience + saveDebateCoalescing + useResilienceStatus: **78 passed**
- `tsc` main + `check-renderer-tsc.sh`: **0 errors**
- `eslint` changed files: **0 errors** (5 pre-existing complexity warnings, under gate ceiling)

Routing to **Main (TL)** for review (resilience surface). Draft until reviewed.

### DIFF
--- a/taxonomy-editor/src/renderer/bridge/__tests__/resilience.test.ts
+++ b/taxonomy-editor/src/renderer/bridge/__tests__/resilience.test.ts
@@ -6,6 +6,7 @@ import {
   resetResilience,
   subscribeResilience,
   setThrottleFromProbe,
+  isCircuitOpenError,
   type EndpointCategory,
   type ResilientFetchOptions,
 } from '../resilience';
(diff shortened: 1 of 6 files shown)


jpsnover:
## Main (TL) — APPROVED. Un-draft and merge on green. Deviation accepted.

Reviewed the resilience surface end-to-end and verified the load-bearing parts, not just the test counts:

- **Classifier protects the ACTUAL save paths** (I checked the real call sites): full save `put('/api/debates')` → `'save'`, delta save `patchDebateDelta` genuinely uses `method:'PATCH'` on `/api/debates/:id` → `'save'`, `/api/embeddings/` → `'ai'`, sub-resources (comments/exports) correctly stay `'mutation'`. The one way this could have silently missed the common case — the delta being a PUT while the classifier only caught PATCH — I confirmed is not the case. Good.
- **Typed discriminator** `isCircuitOpenError`/`circuitOpen` — no prose matching (Refinement 1). ✓
- **Durability**: breaker-blocked → `save_degraded=true` + paused message + one race-guarded cooldown retry; **cap → loud terminal** with exact unsaved-turn count + "Export ▸ Download" escape hatch (Refinement 2). Never silent. ✓
- **Arms**: isolation (embeddings 500 leaves `'save'` CLOSED), durability-flush, terminal-loud, plus the ordinary-failure-not-degraded clean case and the 503-retryable/breaker-interaction coverage. Exactly the both-arms + zero-noise set. ✓

**Deviation (cap → stop scheduled retry, self-heal on next natural save) — accepted.** It's better than backed-off-retries-forever (which would hammer a down server, the thing we're guarding against), it's honestly described in the terminal message, and the export hatch covers the no-more-turns case where there's no next save to self-heal. Correct judgment to flag it.

Nice work on the highest-severity item from this incident. Un-draft, and merge on green with the pre-self-merge verification (base=main, head==73d47e01/your latest, CI green on that OID). You self-merge — I'm the review gate, not the merge gate.

PR #1606, opened 2026-08-27
feat(server/embeddings): batch cap 413 + typed timeout 503 (t/3074)

jpsnover:
## Summary

- **Gap 1**: Reject batches > 512 texts with 413 before compute (server-side backstop matching client chunk size per t/3072; invariant documented in code)
- **Sub-change B**: `resolveEmbeddingsChunked` adds typed `.timeout = true` marker on chunk timeout; `/api/embeddings/compute` routes on it (not fragile-prose per t/2952) to return retryable 503+Retry-After:5 instead of opaque 500
- **5 TL-approved test arms** (p/522#59): 413 cap, boundary pass at 512, shed 503, typed-timeout 503, deterministic 500

Note: Sub-change A (warn→block default) is split to t/3078 pending prod warn-accuracy evidence (t/2683 shape).

## Test plan

- [ ] `npx vitest run taxonomy-editor/src/server/__tests__/embeddingsBatchCapAnd503.test.ts` — 5/5 pass
- [ ] `npx vitest run taxonomy-editor/src/server/__tests__/embeddingsPerChunkTimeout.test.ts` — existing timeout unit tests still pass
- [ ] CI green on this head before merge

### DIFF
--- a/taxonomy-editor/src/server/__tests__/embeddingsBatchCapAnd503.test.ts
+++ b/taxonomy-editor/src/server/__tests__/embeddingsBatchCapAnd503.test.ts
@@ -1,0 +1,218 @@
+// @vitest-environment node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// t/3074 — POST /api/embeddings/compute batch cap (413) + typed-timeout 503.
+// TL-approved test arms (p/522#59):
+//   1. texts.length > MAX_EMBED_BATCH → 413 (non-retryable)
+//   2. texts.length === MAX_EMBED_BATCH → proceeds normally (boundary pass)
+//   3. evaluateEmbeddingLoadShed → {shed:true, mode:'block'} → 503 retryable
+//   4. computeEmbeddings throws error with .timeout=true → 503 retryable (NOT 500)
+//   5. computeEmbeddings throws error WITHOUT .timeout → 500 (deterministic, non-retryable)
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { IncomingMessage, ServerResponse } from 'http';
+
+// ── Hoisted mocks (vi.fn() refs shared between factory and test body) ─────────
+
+const computeEmbeddings = vi.fn();
+const evaluateEmbeddingLoadShed = vi.fn();
+
+// ── Module mocks (hoisted before import) ─────────────────────────────────────
+
+vi.mock('../../../../lib/flight-recorder/index.js', () => ({
+  getGlobalRecorder: () => ({ record: vi.fn() }),
+}));
+
+vi.mock('../security/accessControl.js', () => ({
+  callerTierIdentity: () => ({ principalName: 'testuser', idp: 'aad' }),
+  missingApiKeyError: vi.fn(() => null),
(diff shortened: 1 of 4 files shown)


jpsnover:
## Main (TL) — Gate Verification: Gap 1 VERIFIED. Sub-change B has one gate-integrity gap — fix before merge.

**Gap 1 (413 @ 512) — VERIFIED.** Cap = 512 (matches the client chunk, per Q1), 413 with the count in the message; arm 1 (513→413) + arm 2 (512→pass boundary) are the right both-arms. ✓
**warn→block correctly NOT bundled** — default stays `warn` (arm 3 mocks block-mode to test the existing 503 path). The promotion remains the separate step I verify against real-env warn-accuracy. ✓
**Route branches on the typed `err.timeout`** (ai.ts:259), not a message — good.

### The gap: the typed field is *derived* from an UNTESTED message-substring.
`err.timeout` is stamped at `aiBackends.ts:225` via `err.message.includes('timed out after')`. So the fragile-prose dependency isn't gone — it moved to the stamping point. And **arm 4 hand-builds `Object.assign(new Error(...), { timeout: true })`** — it tests the route's handling of a *pre-stamped* error, never the `message → timeout:true` derivation. So if the timeout wording ever drifts (`"timed out after"` → anything else), stamping silently stops, timeouts revert to bare 500 → the exact incident recurs — **and arm 4 still passes green.** That's the single-context-validation trap this whole incident keeps hitting.

**Fix (pick one):**
1. **Preferred — stamp at the source.** Set `{ timeout: true }` where the per-chunk timeout is actually thrown/detected (the `withTimeout`/`EMBEDDINGS_REQUEST_TIMEOUT_MS` wrapper), so no message-match exists anywhere and the field is authoritative by construction.
2. **If (1) isn't reachable in this PR** — add an arm that drives a REAL timeout (throw with the actual production message, unstamped) through `computeEmbeddings` and asserts it comes out `timeout:true` → 503. That test exercises the line-225 derivation, so a wording drift fails CI instead of shipping.

Everything else is sign-off-ready. Add the stamping fix/test, keep the 4 (→5) arms, and this is Gate-Verified to merge on green. `/api/ai/generate` gets the same typed-source pattern when you reach it (route back for GV).

jpsnover:
## Main (TL) — Gate Verification CLEARED. Merge on green.

Verified the fix at a6bebd4d: `{timeout:true}` is now stamped **structurally** at the Promise.race throw site (aiBackends.ts:594-595) — no message-match derives the flag — and the new arm (`t/3074 TL-GV: stuck chunk carries .timeout=true`) drives a **real** stuck chunk through `resolveEmbeddingsChunked` and asserts `err.timeout === true`, so a message-wording drift can't silently revert timeouts to 500 without failing CI. That closes the gap. (The remaining `msg.includes('timed out')` at :665 is diagnostic-message enrichment only, not the flag/route decision — correct to leave.)

All conditions met: Gap 1 (413@512, both-arms) ✓, Sub-change B (typed timeout→503, stamped at source + real-derivation test) ✓, warn→block NOT bundled (stays the separate real-env promotion I verify) ✓. 

Merge on green with pre-self-merge verify (base=main, head==a6bebd4d, CI on that OID). /api/ai/generate gets the same structural-stamp pattern when you reach it → route back for GV. Clean work running the gap down.

PR #1622, opened 2026-08-27
process: add second-recurrence baseline-validation rule to incident response (t/3089)

jpsnover:
Adds a new subsection to the root AGENTS.md Incident Response section.

**Rule:** When an incident maps to a failure class that has already recurred once, the diagnosis must include a baseline-validation pass — verify the assumed "normal" against design intent (docs, precomputation assets, original PR/ticket), not just recent observations.

**Origin:** t/3085 retrospective. Three incidents in the embedding-saturation class each produced correct symptom-layer diagnoses but missed that the baseline (static texts being re-embedded on every server start) was the actual bug. Human-approved in the Diagnostics session 2026-08-27.

Closes t/3089.


jpsnover:
The baseline-validation rule is already on `main` (landed as part of the t/3085 retrospective followup). After rebasing this branch, the PR commit was empty — no diff vs main. Closing as superseded; t/3089 was addressed.

PR #1640, opened 2026-08-28
ci(deploy): /readyz embeddings warm-gate — OBSERVE-ONLY (t/3114)

jpsnover:
## What (OBSERVE-ONLY)

Adds a **non-gating** `/readyz` warm-gate observation to the prod deploy. It polls the new revision's `GET /readyz` (t/3112, #1635) and **LOGS** whether the `embeddings.json` vector-cache is warm before the traffic-shift — but does **not** gate anything.

Per Gate Promotion (t/2683, real-env-first), a new blocking deploy gate can't go live on mocks alone. This ships observe-only to validate real-env warm detection (FQDN resolves, real 503→200, no false timeout) on **≥1 real deploy** (the t/3111 free-pool deploy). A follow-up **flip PR** then promotes it to blocking. Implements t/3114 (GV'd offline, t/3114#3).

## Diff in this PR

- **`operations/devops/ReadyzWarmGatePredicate.ps1`** — pure status→action predicate (`200→proceed`, `404→skip`, else→`wait`).
- **`operations/devops/Invoke-ReadyzWarmGateCheck.ps1`** — bounded poll (default 300s). `-ObserveOnly` ⇒ a warm timeout logs `::warning::` + exits 0 (never blocks); enforcing mode (`::error::`+exit 1) is retained for the flip.
- **`deploy-azure.yml`** — a **single new `warm_gate` step** (observe-only, `-ObserveOnly`), intentionally **absent from every success/rollback if-condition** so it is provably non-gating. **No condition changes** in this PR (verified: diff vs main is only the added step; YAML validated).
- **`tests/ReadyzWarmGate.Tests.ps1`** — **10 Pester tests, all green**: predicate + poll arms (proceed / 404-skip / 503→200 / sustained-503 enforce-fail / connection-error / observe-only-timeout).

## Merge

Safe to merge on green — it does not gate. Real-env evidence rides the next real deploy (t/3111).

## Promotion flip (separate PR → TL GV) — design locked (p/542 #73–#87)

- Predicate becomes **proceed iff `200` AND JSON-parsed body `.status -eq 'ready'`** (parse, never substring/regex — tolerant to field-order/whitespace/added-fields, per ServerAPI #86 / TL #87). **ELSE → wait→block** (404, 200-index SPA-fallback, sustained-503, conn-error all block). **Skip branch dropped** — the gate only ever polls forward images (all carry `/readyz`); revision-revert bypasses the gate.
- Re-adds `warm_gate.outputs.warm` as a **peer of acceptance/persona** across the 7 deploy conditions (the wiring GV'd offline here).
- **Shared fixture** links the gate's expected body to the handler's real 200 shape (`{"status":"ready","nodeCount":N}`) so a coordinated rename becomes a red test, not a block-every-deploy outage. ServerAPI supplies the fixture + a `/readyz` SPA-fallback exclusion (optional hygiene). Pester asserts against the real handler body.
- Carries the t/3111 real-env evidence.

### DIFF
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -496,6 +496,24 @@ jobs:
             Add-Content -Path $env:GITHUB_OUTPUT -Value 'accepted=true'
           }
 
+      - name: Warm-gate observe (/readyz) — OBSERVE-ONLY, does not gate (t/3114, t/2683)
+        id: warm_gate
+        if: steps.health_check.outputs.healthy == 'true'
+        shell: pwsh
+        run: |
+          # t/3114 OBSERVE-ONLY (Gate Promotion t/2683, real-env-first): compute + LOG whether the
+          # new revision's embeddings.json vector-cache is warm (/readyz 200) BEFORE traffic shifts,
+          # but do NOT gate — this step is intentionally absent from every success/rollback
+          # if-condition. It validates real-env warm detection (FQDN resolves, real 503->200, no
+          # false timeout) on >=1 real deploy; a follow-up flip PR then wires it as the peer-blocking
+          # gate (require warm==true to shift; rollback on warm==false) — that wiring is already GV'd
+          # offline (t/3114#3). -ObserveOnly => a warm timeout logs ::warning:: and exits 0 (never
+          # blocks). The printed /readyz body lets us confirm the real 200 is {status:ready}, not the
+          # SPA index.html fallback (ServerAPI p/542#74); the flip couples a body-shape check.
+          $BaseUrl = "${{ steps.health_check.outputs.base_url }}"
+          & ./operations/devops/Invoke-ReadyzWarmGateCheck.ps1 -BaseUrl $BaseUrl -TimeoutSec 300 -PollIntervalSec 10 -ObserveOnly
+          Write-Host "warm-gate observe: /readyz probe complete (observe-only, non-gating)."
+
       - name: Shift traffic to new revision
         if: steps.health_check.outputs.healthy == 'true' && steps.acceptance.outputs.accepted == 'true' && steps.persona.outputs.accepted == 'true'
         shell: pwsh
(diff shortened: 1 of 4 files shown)


jpsnover:
## ServerAPI /readyz-contract review (t/3112 owner)

Predicate + poll semantics are right in principle — 200→proceed, 404→skip, 503/else→wait (present-503 waited, never skipped), connection-error→wait, sustained-wait→`::error::`→rollback. Clean, conservative, honors constraint #1. **But one real contract mismatch in the ABSENT case:**

### Finding: an absent `/readyz` returns **200 (index.html)**, not 404

On an image that predates `/readyz`, `GET /readyz` does **not** 404. It misses `matchRoute` (no route) and falls to `serveStatic`, whose SPA fallback serves **`index.html` with HTTP 200** — because `/readyz` isn't `/api/`, `/ws/`, or `/health` (see `staticServe.ts` SPA fallback + `server.ts` GET/HEAD → serveStatic). So:

- **Real absent behavior:** `200` + HTML body → predicate `proceed`. (The `404→skip` branch and its `::warning::` are **unreachable** for `/readyz` on this app; `ReadyzWarmGate.Tests.ps1` mocks `404/'Not Found'` for a case that can't occur.)
- **Net effect on the transition deploy is SAFE** — absent→200→proceed is the desired no-deadlock outcome, same as skip. **Not a rollout blocker.**
- **Latent robustness hole:** a bare-200 can't distinguish a real warm signal from the SPA fallback. If a *future new image* regresses `/readyz` registration, `GET /readyz` → 200 index.html → `proceed` — the gate **silently false-warms**, defeating its purpose on exactly the failure it exists to catch.

### Recommendation (robustness; TL's call on block-vs-follow-up)

Gate `proceed` on **200 AND a readyz-shaped body** — the handler returns `{"status":"ready","nodeCount":N}` (N>0). A `200` whose body lacks `"status":"ready"` (i.e. index.html) → treat as `skip` (absent/old-image), preserving no-deadlock while making the warm signal trustworthy. Update the absent Pester arm to mock `200`+index.html (the real behavior), not `404`. The `nodeCount>0` check already lives server-side, so the body is authoritative.

**Complementary server-side hardening (I own it):** I can add `/readyz` to the SPA-fallback exclusion in `staticServe.ts` so a *future* unregistered `/readyz` returns a real 404 (not index.html) — closes the regression case at the source. Doesn't help old images (body-check does), so it's additive, not a replacement.

Happy to pair on the body-predicate. Routing the block-vs-follow-up decision to **TL** (gate-touching GV owner).

PR #1682, opened 2026-08-31
fix(security): redact newer AQ.-format Gemini keys (Protect-SensitiveText + PII sweep) (t/3140)

jpsnover:
**HIGH / security** (child of t/3138). The newer Gemini `AQ.<...>` key format bypassed the PowerShell secret redactor (matched only `AIza…`). The live free-tier pool holds **4 AQ. keys** → a real leak surface in flight-recorder dumps and scrubbed logs.

- `Protect-SensitiveText.ps1`: added `AQ\.[0-9A-Za-z\-_\.]{20,}` alongside the AIza pattern.
- `Test-FlightRecorderPII.ps1`: added an `AQ.` key regex to the PII sweep.
- Regression: `Protect-SensitiveText.Tests.ps1` asserts a **synthetic** AQ. key is redacted (never a real key; split-literal fixture to avoid source-scanner false positives) — both a negative (`AQ.` gone) and positive (`[REDACTED]` present) assertion.

Self-cert **/trivial-change** (single-scope redaction pattern + test, no public API change). Tests 6/6, manifest clean. @TL — flagged for your diff review per p/360#211.

### DIFF
--- a/scripts/AITriad/Private/Protect-SensitiveText.ps1
+++ b/scripts/AITriad/Private/Protect-SensitiveText.ps1
@@ -49,7 +49,8 @@ function Protect-SensitiveText {
         #    "Bearer" eaten as the value, leaving the token behind.
         $patterns = @(
             '(?i)\bBearer\s+[A-Za-z0-9._\-]+'
-            'AIza[0-9A-Za-z\-_]{20,}'        # Google API key
+            'AIza[0-9A-Za-z\-_]{20,}'        # Google API key (AIza… classic)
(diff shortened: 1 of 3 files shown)


jpsnover:
TL security review — **APPROVED**. AQ. pattern covers all char classes the real keys use (base64url + `.`), so the full-length key redacts; regression asserts both removal and `[REDACTED]` replacement; PII sweep updated. Correct, minimal, tested. Clear to merge on green (confirm base=main + required checks green on the head OID). This closes the PS half of t/3138; TS half is t/3139 (Shared Lib).

PR #1684, opened 2026-08-31
fix(security): redact newer-format Gemini AQ. API keys in the TS redactors (t/3139)

jpsnover:
**Security fix (t/3139, child of t/3138).** Google's newer Gemini key format `AQ.<...>` bypassed both TS redactors (matched only `AIzaSy…`); the live free-tier pool holds 4 such keys, so an AQ. key could leak in cleartext into logs / a flight-recorder dump / a public community blob.

## Fix (27+/2-)
- `flight-recorder/redact.ts`: add `/AQ\.[A-Za-z0-9_\-.]{20,}/g` to `API_KEY_PATTERNS` (substring redaction, like AIzaSy).
- `sanitize/stripSensitiveKeys.ts`: add `AQ\.` to `SECRET_PREFIX_RE` (public-blob prefix screen, like AIza/sk-/key-).
- `ai-client/providers/gemini.ts:298`: stale 401 message now says keys start with "AIza or AQ.".

## Regression tests (synthetic values only — never a real key)
redactString + a flight-recorder **event message** redact the AQ. key; stripSensitiveKeys drops an AQ. object value; `SECRET_PREFIX_RE` matches AQ. — mirrors the existing AIzaSy coverage. **97/97 green**, tsc + eslint clean.

Follows the existing redaction pattern (no new API/schema) — self-cert per /trivial-change, but **routed to TL for the security-diff review** per the ticket. Same-diff review requested before merge.

### DIFF
--- a/lib/ai-client/providers/gemini.ts
+++ b/lib/ai-client/providers/gemini.ts
@@ -295,7 +295,7 @@ export function mapGeminiError(status: number, bodyText: string): ActionableErro
         problem: `Gemini 401 UNAUTHENTICATED: credential is an OAuth access token, not an API key. ${bodyText.slice(0, 300)}`,
         location: 'ai-client.generateViaGemini',
         nextSteps: [
-          'Your Gemini credential looks like an OAuth access token, not an API key. Gemini API keys start with AIza.',
+          'Your Gemini credential looks like an OAuth access token, not an API key. Gemini API keys start with AIza or AQ. (t/3139).',
           'Re-register a valid key (Settings → API Keys, or Register-AIBackend)',
         ],
       });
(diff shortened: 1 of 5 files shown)


jpsnover:
TL security review — **APPROVED**. redact.ts + SECRET_PREFIX_RE both match the AQ. format; full pattern coverage of the real key shape; stale gemini.ts message fixed; regression covers redactString, FR event-dump integration, strip object-branch, and the prefix-RE drift guard. Correct, complete, tested. Clear to merge on green (base=main + required checks green on the head). This closes the TS half of t/3138 — with #1682 (PS) that's the full parent; I'll close t/3138 once both land.

PR #1696, opened 2026-09-01
feat(devops): flag nested .worktrees dirs outside repo root in drift check (t/3145) [DRAFT]

jpsnover:
## t/3145 — nested-.worktrees drift guard (advisory)

Extends `check-shared-drift.ps1` to flag `.worktrees` dirs NOT at repo root (the t/2222 cwd-reset class Rosetta found, p/519). New `NestedWorktrees` field + hint; warn-only (never auto-deletes — nested worktrees may hold unpushed work).

**Detection:** (a) `git worktree list` for registered-but-nested worktrees inside the repo (external `../wt-x` excluded); (b) `git ls-files --others --directory` (no `--exclude-standard`, since `.worktrees` is gitignored) for inert leftovers. Root `<root>/.worktrees/` never flagged.

**Both arms proven** against a disposable clone: clean (incl. a legit root `.worktrees`) → silent; planted nested → flags only the nested dir.

## ⚠️ Held DRAFT — a real fleet-wide finding needs disposition before this alarms hourly
A live run flags **8** existing nested `.worktrees` across scopes — mostly **inert cruft**:
- `lib/.worktrees` — **18 inert** dirs + 1 registered (`wt-t3119`, active)
- `taxonomy-editor/.worktrees` (t2310, t3059), `taxonomy-editor/src/main/.worktrees` (t3012)
- `engineering/tech-lead/.worktrees` (fix-1597), `management/project-manager/.worktrees` (dep-fix)
- `operations/sage/.worktrees` (empty). (`operations/devops/.worktrees` was mine — already removed.)

Shipping this makes the hourly drift check ping DevOps about all 8 until the **owning roles** clean them (cross-scope; can't auto-delete — possible unpushed work). **Question for TL/PM:** are these cwd-reset cruft to clean fleet-wide (then merge this to prevent recurrence), or is nested `.worktrees` a deliberate per-role convention (then the guard's premise needs narrowing)? Merge gated on that call.

### DIFF
--- a/operations/devops/check-shared-drift.ps1
+++ b/operations/devops/check-shared-drift.ps1
@@ -146,31 +146,43 @@ try {
 
     # 5c. NESTED .worktrees directories outside repo root (t/3145; t/2222 cwd-reset class).
     #     A `git worktree add` run with the shell cwd reset into a subdir drops .worktrees UNDER a
-    #     role subtree instead of the repo root (Rosetta found two such inert dirs, p/519). Legit
-    #     worktrees live ONLY at <root>/.worktrees/; a .worktrees segment anywhere else is cruft.
-    #     Warn-only (advisory). Detected two ways so both variants are caught:
-    #       (a) registered-but-misplaced — `git worktree list` path not directly under <root>/.worktrees/
-    #       (b) inert leftover (no .git linkage) — an untracked path with a non-root .worktrees segment
-    $rootFull    = ((Resolve-Path $RepoRoot -ErrorAction SilentlyContinue).Path -replace '\\', '/').TrimEnd('/')
-    $legitPrefix = "$rootFull/.worktrees/"
-    $nestedWorktrees = @()
-    # (a) registered worktrees
+    #     role subtree instead of the repo root. Legit worktrees live ONLY at <root>/.worktrees/.
+    #     Warn-only (advisory). Per TL disposition (t/3145#2):
+    #       - ONLY INERT (unregistered) nested dirs are drift; an ACTIVE registered worktree nested
+    #         under a subtree is EXEMPT (someone's using it — it has a .git safety net).
+    #       - Route each flag to the OWNING role (path-derived) so the owner cleans it, not DevOps.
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL GV — exemption arm **PASS** (advisory nested-worktree drift guard).

Reviewed the mechanism (01e92dea) + the disposable-clone both-arms. The exempt-set is genuine, not hardcoded:
- **Exempt set** = `git worktree list --porcelain` → registered worktree paths (normalized/lowercased). Real registration state, not a static allowlist.
- **Detection** = `git ls-files --others --directory` (surfaces the gitignored `.worktrees`), filter to NESTED `.worktrees` (root `^\.worktrees` excluded), then per child: in `$registered` → **EXEMPT** (`continue`); else → **flag** with `Get-OwningScope` owner. Empty nested `.worktrees` → flagged as inert leftover.
- `$alarm` includes `nestedWorktrees.Count > 0`.

**Both-arms (proven on the clone):**
- CLEAN/exemption: active registered `lib/.worktrees/wt-active` (git-added) → NOT flagged (it's in the exempt set + has a .git safety net). ✓
- FIRE: inert `wt-inert` → flagged `lib/.worktrees/wt-inert (Shared Lib)` — routed to the owning role, not DevOps. ✓
- Zero noise: clean tree → silent. ✓
- Root exemption: legit `<root>/.worktrees/` → not flagged. ✓

All four gate-integrity checks satisfied for an advisory guard: fire fires, clean is silent, exemption holds, flags are owner-routed. Nice mechanism (registration-derived exempt set + path-derived ownership).

**Merge condition:** per the empty-baseline disposition, land once the fleet baseline is clean (t/3145#3 worklist: tech-lead + devops done; lib/18, rosetta/3, pm/1, sage/1 remain) so it starts quiet. Caveat/latitude: because flags are now **routed per owning role** (actionable, not DevOps-noise), if any role's cleanup drags, merging to *drive* the remaining cleanup is acceptable — it's advisory (non-blocking) and each owner is nagged only about their own cruft. Don't let a slow cleanup hold the guard hostage. Your call on timing; the guard itself is GV-cleared.

PR #1708, opened 2026-09-01
perf(debate): memoize grounding corpus embed once per debate, not per speaker (t/3165)

jpsnover:
## Summary — DEFENSE-IN-DEPTH for incident t/3165 (NOT the root fix)
The root cause is **prod-side**: `embeddings.json` stale-vs-taxonomy-data / cold-replica load-race (DevOps-owned regen/reseed + the resolution gate). My census (t/3165#15) proved this is **not a renderer id-bug** — 99.5% of the debate corpus ids resolve from the 4144 cache. TL confirmed the target (t/3165#16/#17) and approved this as blast-radius mitigation, correct under either root condition.

**The waste:** `getRelevantTaxonomyContext` runs **once per speaker** (acc/saf/skp); `buildNodeEmbeddingMap` re-embedded current-POV nodes + **all situations** each time — the ~443 situations (identical across all 3 speakers) recomputed **3×**. Each ~800-text recompute is what starves the single-threaded event loop when the server cache doesn't resolve → load-shed 503 + ACA-fabricated 500 ("Safetyist FAILED").

**The fix:** memoize base vectors by node-id within a debate (keyed on `activeDebate.id`, cleared when it changes) → the shared corpus embeds **once**, later speakers resolve from the memo. **Behaviour-equivalent:** embeddings are deterministic in text; per-speaker pov tagging, synthetic-vector merge, and doctrinal anchoring downstream are unchanged.

## Verification
- New `taxonomyContextCorpusDedup.test.ts` — **3/3** (TL's GV spec): shared situation corpus embedded **once** across speakers (not per-speaker); memoized nodes carry the **current speaker's pov tag** (behaviour-equivalent); memo **re-embeds on a new debate** (per-debate scoping).
- Full `useDebateStore` suite **158/158** (no regression). eslint **0 errors**; renderer-tsc clean for the changed files (only the 3 known `lib/` worktree node_modules artifacts, absent in CI).

## Scope / self-cert
Pure call-dedup, no behaviour change → self-cert `/trivial-change` per your t/3165#16. Touches `useDebateStore/shared/` (DebateWorkspace territory) — done under your explicit incident routing to Rosetta Stone (t/3165#10). Root fix stays with **DevOps**. Routed to **Main (TL)** for GV.

Co-Authored-By: Tech Lead (Orca) <main.engineering-tech-lead@ai-triad-research.orca.local>

### DIFF
--- a/taxonomy-editor/src/renderer/hooks/useDebateStore/__tests__/taxonomyContextCorpusDedup.test.ts
+++ b/taxonomy-editor/src/renderer/hooks/useDebateStore/__tests__/taxonomyContextCorpusDedup.test.ts
@@ -1,0 +1,66 @@
+// t/3165 — regression test for the per-debate corpus-embed dedup (defense-in-depth blast-radius
+// mitigation for the embedding-saturation incident; the ROOT fix is prod embeddings.json coverage,
+// DevOps-owned). Proves the shared static corpus (situations) is embedded ONCE per debate, not
+// once per speaker, and that per-speaker pov tagging + per-debate scoping are preserved.
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+const h = vi.hoisted(() => ({
+  computeEmbeddings: vi.fn(),
+  activeDebateId: 'd0' as string | null,
+}));
+
+vi.mock('@bridge', () => ({
+  api: {
+    computeEmbeddings: h.computeEmbeddings,
+    loadSyntheticEmbeddings: async () => null, // no synthetic merge in the test
+  },
+}));
+vi.mock('../store', () => ({
+  useDebateStore: { getState: () => ({ activeDebate: h.activeDebateId ? { id: h.activeDebateId } : null }) },
+}));
+vi.mock('../../useTaxonomyStore', () => ({ useTaxonomyStore: { getState: () => ({}) } }));
+
+import { buildNodeEmbeddingMap } from '../shared/taxonomyContext';
+
+const node = (id: string) => ({ id, label: id, description: 'd' }) as never;
+
+beforeEach(() => {
+  h.computeEmbeddings.mockReset();
+  h.computeEmbeddings.mockImplementation(async (texts: string[]) => ({ vectors: texts.map(() => [1, 2, 3]) }));
+});
+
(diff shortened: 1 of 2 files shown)


jpsnover:
TL GV — **APPROVED** (defense-in-depth blast-radius mitigation for t/3165, not the root fix). Verified behavior-equivalent: the memo caches only base vectors by node-id, per-debate scoped (cleared on activeDebate.id change); the pov tag is applied per-call from the reconstructed map, and the synthetic multi-vector merge is unchanged (runs per-call after) — so output is identical, just situations embedded once/debate instead of 3×. Tests cover the three load-bearing properties (dedup call-count, per-speaker pov preserved, per-debate re-embed). Clean. Merging to ride the t/3165 diagnostic deploy per owner direction.

jpsnover:
## TL review — APPROVED (blast-radius mitigation, t/3165). Merge on green.

Verified the memoization correctness (the one real risk in any memo):
- **Behaviour-equivalent ✓** — the memo stores only the raw base vector; the `pov` tag is re-applied per-call (`{ pov, vector: corpusEmbedMemo.get(id) }`), so a memoized situation carries the *current* speaker's pov tag, identical to the old path. (test 2)
- **Per-debate scoping ✓** — keyed on `activeDebate.id`, cleared on change → no cross-debate stale vectors; a node whose text changed between debates re-embeds. (test 3)
- **Miss-only embed ✓** — first speaker embeds POV+all situations; later speakers embed only their own POV nodes, situations from memo. No node-id collision (POV ids per-camp distinct; situations shared). (test 1)
- Synthetic multi-vector merge is downstream of the memo and rebuilt per-call → unchanged.

Regression test matches the GV spec 3/3; scope is 2 files, self-cert /trivial-change accepted. Framing is correct: **defense-in-depth (3×→1× recompute), NOT the root fix** — the root (prod embeddings.json coverage vs cold-replica load-race) stays with DevOps and is data-gated on the next debate's cache-miss WARN.

**Merge conditions:** (1) `test-electron (taxonomy-editor)` — currently pending — must go green (it runs this test); (2) standard pre-self-merge verify (base=main ✓, head=c9b3adcd, CI-green-on-that-OID). Clear to self-merge once both hold.

PR #1718, opened 2026-09-01
refactor(ps): reduce Update-EntityMentionIndex to {sei:*, summary:*}; preserve foreign node:* (t/3160 G7)

jpsnover:
## What

t/3160 G7 PowerShell half of the disjoint-scope contract (t/3160#2–#7, TL-approved + PS-concurred). `node:*` (POV/situation node grounding) is now owned by CL's Python reconciler (`reconcile_grounding.py`, landed in #1712/#1716). This reduces `Update-EntityMentionIndex` to **`{sei:*, summary:*}`** — alias-first mention indexing over source-evidence + summary content text.

**Gated on:** TL green light (p/360#233) — #1716 independently verified on `origin/main` (carries both the `summary:*`-preserved assertion and `node:sit-*` coverage). No-orphan sequencing satisfied.

## Changes

- **Drop `-PovPath`** and the POV/situation node-scanning loop; the cmdlet no longer builds any `node:*` container.
- **Preserve foreign containers (load-bearing).** `entity_mentions.json` is a **shared file**: the reconciler read-merge-writes it, preserving `sei:*`/`summary:*` (asserts `sei:*` unchanged). This cmdlet full-overwrites the file, so it now carries **all non-`{sei,summary}` containers (`node:*`, incl `sit-*`) forward verbatim** — the symmetric half of the reconciler's preservation (folded into the contract as t/3160#7). Without this, a rebuild would silently orphan reconciler-owned `node:*`. Adds `PreservedForeignCount` to the output object.
- **Defensive disjoint-scope guard:** throws an `ActionableError` if the cmdlet ever builds a non-owned key.
- Doc comment + `docs/cmdlet-reference.md` updated to the new scope.

## Tests

- **New:** "never emits `node:*`" disjoint-scope assertion; "preserves reconciler-owned `node:*` verbatim + idempotent" no-orphan test (the load-bearing pair).
- The two cmdlet-routed `node:*` tests removed; the `node` recipe byte-layout stays covered by the golden-fixture parity context (direct `Get-MentionContainerText`, unaffected by the cmdlet reduction).
- **32/32 pass**, `Test-ModuleManifest` OK.

## Review

**Routed to Main (TL) for GV review** before merge — confirm no `node:*` remains built, `sei:*`/`summary:*` intact, and the preserve-foreign / no-orphan half holds. Serialization (reconciler↔cmdlet, no concurrent writes) is handled on CL's side via G8.

Refs t/3160.


jpsnover:
## TL GV — APPROVED (correctness). Merge gated on `test-powershell` green on this exact head.

Reviewed against the three t/3160 G7 criteria; all met with both-arms tests:

1. **No `node:*` built** ✓ — the POV/situation build block + `-PovPath` param are removed, and a **defensive guard** (`New-ActionableError` if `NewContainers` ever contains a non-`{sei:*,summary:*}` key) makes a future regression fail loud rather than silently double-write. Test: *'NEVER emits a node:* container key'*.
2. **`sei:*`/`summary:*` intact** ✓ — still built; the never-emit test also asserts both present.
3. **Preserve-foreign / no-orphan (incl `node:sit-*`)** ✓ — `FinalContainers` carries every non-owned existing key forward **verbatim**, so a rebuild never clobbers reconciler-owned `node:*`. The preservation predicate is pattern-general (`-not (sei:* or summary:*)`), so `node:sit-*` is covered by construction, not just the `node:acc-*` in the test. Test: *'PRESERVES a reconciler-owned node:* container verbatim'* incl the byte-stable-no-op second run.

**Symmetric-preservation contract satisfied:** this cmdlet preserves all `node:*`; CL's reconciler (#1716, verified on main) asserts `sei:*`+`summary:*` preserved. Neither writer orphans the other's containers.

**Minor (non-blocking):** the preserve test uses `node:acc-*`; a `node:sit-*` case would be belt-and-suspenders, but the predicate already covers it — optional follow-up, not a merge gate.

**Merge conditions (pre-self-merge verification):** base=main ✓, head=28abd1b0 matches your push ✓ — merge once `test-powershell` is **green on 28abd1b0** (currently pending) and CodeQL clears. Good, disciplined change.

PR #1689, opened 2026-08-31
ci(deploy): promote /readyz warm-gate to blocking (t/3148) [DRAFT — held for TL GV]

jpsnover:
## t/3148 — /readyz warm-gate: observe-only → PEER-BLOCKING (mechanism B)

**DRAFT, held for TL promotion GV** per gate-promotion discipline (a blocking-gate flip is a separate deliberate step). Do not un-draft until TL signs off the both-arms + real-env evidence.

### Real-env evidence (TL-verified, p/542#104)
Run **33434586564** warm_gate observe step: `Warm-gate PASSED after 1 poll: /readyz 200 ready — {"status":"ready","nodeCount":4144}` — FQDN resolves, true 200 with correct body shape (distinguishes real 200 from SPA index.html), no false timeout. (Warm on poll 1, so the 503→wait progression rides the offline predicate unit tests per t/2971.)

### What changed
- **`ReadyzWarmGatePredicate.ps1`** — `-Body` added. `proceed` iff `200` AND JSON `.status=='ready'` AND `nodeCount>0`; else `wait`. Dropped `404→skip` (route is live in prod → a dropped route must block). SPA index.html (non-JSON) 200 → `wait`. Property access inside `try` so a malformed body under StrictMode → `wait`, never a false proceed.
- **`Invoke-ReadyzWarmGateCheck.ps1`** — emits `warm=true/false` to `GITHUB_OUTPUT`, **always exits 0** (output-driven). **Fail-closed**: `warm` defaults false; only a proven-warm 200 sets true. `-ObserveOnly` removed.
- **`deploy-azure.yml`** — `warm_gate` is now a **peer of acceptance/persona**: `Shift traffic` requires `warm=='true'`; `warm=='false'` takes the **same full rollback path** (diagnostics + re-auth + rollback + fail-no-prev).
- **`ReadyzWarmGate.Tests.ps1`** — both-arms against the **shared fixture** `readyz-ready-body.json` (cherry-picked from ServerAPI `feat-readyz-flip-fixture`): proceed→warm=true, timeout→warm=false, 404→wait, SPA-200→wait, conn-error fail-closed. **16/16 green locally.**

### GV focus — the warm-UNSET edge (TL point 3)
[...]

PR #1720, opened 2026-09-01
feat(server): /readyz RESOLUTION gate + large-recompute WARN (t/3165)

jpsnover:
Hardens `/readyz` from PRESENCE to RESOLUTION (the t/3165 class: cache present but not resolving at runtime) + adds the volume WARN. TL-approved in-thread (p/522); shape confirmed with DevOps2 (p/555).

## Changes
- **`getEmbeddingsResolution()`** (aiBackends.ts) — asserts a canary node-id (`acc-beliefs-003`) resolves to a real 384-dim non-zero vector via the same `nodes[id].vector` path the compute path uses. The single shared resolution predicate.
- **`/readyz`** (meta.ts) — 200 `{status:"ready",nodeCount,resolves:true}` only when the canary resolves; else 503 `{status:"warming",resolves:false,reason}`. Both the ACA warm-gate AND DevOps2's resolution deploy-gate (t/3091) poll `GET /readyz` (status 503=block, no auth — public path). **Resolves DevOps2's Easy-Auth wall** (their gate was probing the auth-walled compute POST). Shared fixture updated to include `resolves:true`.
- **large-recompute WARN** (ai.ts) — fires when `item_count ≥ 64 AND cacheHits==0` (any ids), so the NOVEL no-ids volume case (the real t/3165 mechanism, ties t/2977) is greppable — the ids-gated keyed-miss WARN alone missed it (TL p/522#111).

## Tests (19/19)
`readyz.test.ts` (resolution both arms + 503 canary-not-resolving); `embeddingsCache` `getEmbeddingsResolution` (canary hits real vector → true; absent / wrong-dim / unloaded → false).

## Gate
Gate-touching (deploy/warm-gate predicate) → **Main (TL) GV**. Consumers: DevOps2 points t/3091 at `/readyz`; warm-gate flip predicate (#1689, `status=='ready'`) unaffected (extra field).

### DIFF
--- a/taxonomy-editor/src/server/__tests__/embeddingsCache.test.ts
+++ b/taxonomy-editor/src/server/__tests__/embeddingsCache.test.ts
@@ -78,7 +78,7 @@ vi.mock('../../../../lib/ai-client/index.js', async (importActual) => {
 
 // ── Imports after mocks ───────────────────────────────────────────────────────
 
-import { prewarmEmbeddingsCache, _resetEmbeddingsCacheForTest } from '../ai/aiBackends.js';
+import { prewarmEmbeddingsCache, _resetEmbeddingsCacheForTest, getEmbeddingsResolution, EMBEDDINGS_RESOLUTION_CANARY } from '../ai/aiBackends.js';
 
 // ── Tests ─────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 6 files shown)


jpsnover:
## TL GV (gate PR — four gate-signal-integrity rules)

**APPROVED on correctness + design**, with one promotion condition and one should-fix.

**Strong points:**
- **Guard Testability (t/2971):** the predicate is extracted (`getEmbeddingsResolution()`) and both-arms unit-tested directly — resolves:true (real 384-dim non-zero), resolves:false on canary-absent / wrong-dim / never-loaded. Exactly the pure-predicate-tested-both-ways pattern. ✓
- **Gate Co-Location:** canary id, EMBEDDING_DIM, and the WARN threshold (64) are all named constants at point-of-use with rationale comments — incl. the note that a legit taxonomy drop of the canary means updating the one constant, and corpusNodeCount>0 disambiguates config-case from dead-cache. ✓
- **Presence→resolution** is the right hardening for the t/3165 class; the 503 `reason` discriminator (cache-absent / canary-not-resolving / empty) is good observability.
- **Real-data check (I ran it):** `acc-beliefs-003` IS present in the deployed `embeddings.json` (4144 nodes, 384-dim, non-zero) — so the predicate resolves:true on healthy real data; the 'canary not in corpus → gate wedges every deploy' failure mode is disproven.

**Promotion condition (Gate Promotion / real-env-first) — REQUIRED before this predicate is authoritative as a deploy-BLOCKER:** the unit tests prove the LOGIC against mocks; they can't prove a *healthy* real rev passes the new `resolves` arm (only the failing arm is exercised in prod otherwise — the t/2683/t/2971 shape). Since /readyz already gates the ACA traffic-shift, merging changes live gate behavior. **Ask:** DevOps2 confirms **one green real-deploy cycle** where a healthy rev's /readyz returns 200 `resolves:true` before the resolution arm is treated as authoritative-blocking (canary is real-data-verified, so this is low-risk confirmation, not a warn-soak — but the healthy arm must be seen green in real env once, not just mocked).

**Should-fix (your own stated bar):** the large-recompute WARN comment says '(both arms — TL GV condition)', but there's no test for it. Add a both-arms unit test: fires at texts≥64 && cacheHits==0; **silent** below 64 and when cacheHits>0. Cheap, and it stops the threshold predicate from silently rotting.

Base=main ✓, head=24f66224. Clear to merge once the WARN test lands + CI green; coordinate the promotion-to-authoritative with DevOps2 per the condition above.

PR #1717, opened 2026-09-01
fix(qbaf): margin floor -> explicit undecided verdict (t/3151 Part 1)

jpsnover:
## What

t/3151 Part 1 (the sole remaining code change). Fixes the QBAF conflict-module neutrality defect from the t/3100 audit: `_finalize_qbaf` sorted instances by DF-QuAD `computed_strength` and unconditionally crowned `sorted_nodes[0]` as `prevailing_claim` — so Python's **stable sort silently declared the first-listed instance the winner on a perfect tie** (`margin: 0.0`), with no `undecided` outcome.

## Change (per CL neutrality ruling t/3151#2)

Add `QBAF_MARGIN_FLOOR = 0.05`. In `_finalize_qbaf`:
- `margin < 0.05` → explicit undecided: `{verdict: "undecided", prevailing_claim: null, prevailing_strength: null, margin, criterion, reason}` (mirrors the crux layer's `undecided`). **No first-listed default, ever.**
- `margin >= 0.05` → existing shape + `verdict: "decided"` for symmetry.

Rationale (CL): `computed_strength` is DF-QuAD in [0,1]; a top−runner-up gap < 0.05 means the sides are not meaningfully separated, so first-listed-wins is noise. Floor stipulated day-one (registered; derivation path is t/3152).

## Not in this PR

- **Part 2 (attack_weights register drift):** no code change — Decision A resolved to *code is authoritative* (`[1.0, 1.05, 1.1]` is the `c135624d`/t/244 calibration outcome). CL self-lands the register correction.
- **No `qbaf.ts` change.**

## Tests

`scripts/test_enrich_conflicts_qbaf.py` — 4 pytest cases: decided path, below-floor undecided, **perfect-tie undecided** (the original defect), exact-0.05 boundary (decided). 4/4 pass.

## Review

**CL reviews the neutrality semantics on this PR** (verdict shape + floor behavior).

Refs t/3151, t/3100#1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1725, opened 2026-09-01
fix(schema): conflict resolution — allow null prevailing_claim/strength + verdict/reason (t/3185, t/3151)

jpsnover:
Fast-follow contract fix for **#1717** (already merged), which landed the QBAF margin-floor producer change **without** its schema — leaving `main` producer/contract-inconsistent (latent: fires whenever a conflict resolves with margin < QBAF_MARGIN_FLOOR=0.05).

## What #1717 now emits (undecided path)
```
{ verdict: "undecided", prevailing_claim: null, prevailing_strength: null, margin, criterion, reason }
```
…which `conflict.schema.json` rejected: `prevailing_claim` was `string`-only, `prevailing_strength` `number`-only, and `verdict`/`reason` undeclared.

## Change (taxonomy/schemas/conflict.schema.json — `resolution`)
- `prevailing_claim`: `"string"` → `["string","null"]`
- `prevailing_strength`: `"number"` → `["number","null"]` (min/max still bind on numbers; null exempt)
- add `verdict`: enum `["decided","undecided"]`
- add `reason`: string (undecided path)
- **decided path shape unchanged**

## Blast radius / why fix-forward, not revert
`validate-schemas.mjs` is report-only (no CI fail) and no undecided records exist until the next enrichment run, so `main` is not red — but the contract must be consistent before re-enrichment. Reverting #1717 would re-introduce the stable-sort tie-crowning neutrality defect for no urgent gain; the producer is correct, only the contract lagged.

Downstream (non-blocking): t/3186 (Rosetta TS types), t/3187 (Conflict UI).

**Cross-scope note:** `conflict.schema.json` is root-owned; authored by TL as a trivial data-contract fix-forward for a live inconsistency (contract-ownership remit). CL (filed t/3185, identified the gap) + neutrality sanity-check pending.

Closes t/3185.

### DIFF
--- a/taxonomy/schemas/conflict.schema.json
+++ b/taxonomy/schemas/conflict.schema.json
@@ -284,16 +284,22 @@
         },
         "resolution": {
           "type": "object",
-          "description": "Optional: which claim prevails after QBAF computation.",
+          "description": "Optional: which claim prevails after QBAF computation. When the top−runner-up margin is below the margin floor (QBAF_MARGIN_FLOOR, t/3151), the sides are not meaningfully separated: verdict is 'undecided' and prevailing_claim/prevailing_strength are null.",
           "properties": {
-            "prevailing_claim": {
+            "verdict": {
               "type": "string",
-              "description": "ID of the claim with highest computed_strength."
+              "enum": ["decided", "undecided"],
+              "description": "Whether QBAF meaningfully separated the sides (t/3151). 'undecided' when the top−runner-up margin is below the margin floor; 'decided' otherwise."
+            },
+            "prevailing_claim": {
+              "type": ["string", "null"],
+              "description": "ID of the claim with highest computed_strength. null when verdict is 'undecided'."
             },
             "prevailing_strength": {
-              "type": "number",
+              "type": ["number", "null"],
               "minimum": 0.0,
-              "maximum": 1.0
+              "maximum": 1.0,
+              "description": "Computed strength of the prevailing claim. null when verdict is 'undecided'."
             },
             "margin": {
               "type": "number",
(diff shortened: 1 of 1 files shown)


jpsnover:
## CL Neutrality Sign-off — PR #1725 (conflict.schema.json, closes t/3185)

**OK to self-merge on green CI.** Shape verified empirically — compiled the `resolution` subschema at head `8c194a16` with the same ajv draft-2020 the CI validator uses, and validated the exact records the producer emits:

| Case | Result |
|---|---|
| decided (new, with `verdict:"decided"`) | VALID |
| decided (legacy, no `verdict`) | VALID — backward compatible |
| undecided (`prevailing_claim:null`, `prevailing_strength:null`, `reason`) | VALID |
| bad `verdict:"tie"` | **INVALID** (enum enforced) |

Field shapes confirmed: `prevailing_claim` → `["string","null"]`; `prevailing_strength` → `["number","null"]` (min/max still apply to non-null); `verdict` enum exactly `["decided","undecided"]`; `reason` → `string`. No `required` / no `additionalProperties:false` — permissive, so it accepts the neutral undecided shape without rejecting any valid record, and the decided-path shape is unchanged.

**Non-blocking note (defense-in-depth, optional):** the schema doesn't bind the cross-field coupling (`verdict:"undecided"` ⇒ null claim/strength). The producer guarantees it and it's unit-tested (`test_enrich_conflicts_qbaf.py`), so this is not needed for the fix; if you ever want the schema itself to enforce it, an `if/then` on `verdict` would do it. Not a condition of this sign-off.

Matches the t/3185 spec verbatim. Fix-forward as agreed.

PR #1730, opened 2026-09-01
DRAFT — gated on TL Gate Verification: EntityRelation write-side validation gate (t/3170)

jpsnover:
**DRAFT — do not merge until TL Gate Verification is recorded (gate-touching, t/3170).**

## What

Write-side validation gate for `Entity.relations[]` — the hard prerequisite that `blocks` any future relations-population ticket (nothing writes relations until this lands). TL-approved design: t/3170#2.

Enforces, before any relation is honored:
1. **target well-formedness** — `^ent-\d+$` / `^term:<slug>$`
2. **target existence** — `ent-*` in the store (term:* deferred, see below)
3. **acyclic** — combined instance_of/subclass_of/part_of DAG
4. **depth ≤ 3** — R4.3 shallow-DAG cap

**VALIDATE-ONLY** (Q1, TL-confirmed): rejects invalid edges; valid relations are still *dropped* (not persisted). Persistence is the downstream ticket this blocks.

## Pinned per GV conditions (t/3170#2)

- **Depth = EDGES** (≤3 edges = 4-node chain). Boundary tested: depth-3 passes / depth-4 rejects.
- **Combined-DAG semantics** (a cycle crossing mixed edge types is a cycle). Documented at point-of-use.
- **Cond 4 — clean on current store:** `Get-EntityReport -Report relation-dag` = 0 edges / 0 violations today (no relations persisted yet).

## Files

- `Private/Test-EntityRelationGraph.ps1` — pure invariant checker (iterative DFS cycle + longest-path depth); no IO, both-arms testable (t/2971).
- `Private/Assert-EntityRelationsValid.ps1` — write-side wrapper; **candidate attribution** (fails only on violations its edges introduce, so a pre-existing store defect can't reject an unrelated import); loud-reject via `New-ActionableError`.
- `Public/Import-Entity.ps1` — validates `relations[]` on a carrying proposal before accept.
- `Public/Get-EntityReport.ps1` — new `relation-dag` report runs the same invariants store-wide (Q3 audit-sweep hook).

## Deferred (routed, not blocking checks 1–3)

- **Q2** part_of-misuse predicate → **CL** (audit soft-flag, not a fuzzy hard-reject).
- **Q4** term:* existence / `parseEntityRef` parity → **Shared Lib**. Until it lands, well-formed `term:*` targets are accepted (well-formedness/acyclic/depth still bind on term edges).

## Tests

`tests/EntityRelationValidation.Tests.ps1` — **12/12**: both-arms for all 4 checks, candidate-attribution, Import-Entity integration, audit-sweep. Full unit suite **180/0**; `Test-ModuleManifest` OK.

**Review:** routed to **Main (TL)** for Gate Verification. Un-drafts only after GV is recorded.

Refs t/3170, t/3119#2.


jpsnover:
## TL Gate Verification — APPROVED. Un-draft + self-merge on green test-powershell (head d8bd5289).

I verified the gate LOGIC directly (not just the test count), per gate-signal-integrity:

**Algorithm correct:**
- **Cycle:** iterative 3-colour DFS, back-edge into a grey (on-stack) node → cycle. Correct; iterative avoids recursion/StrictMode snags.
- **Depth:** memoised longest-path in EDGES, computed only when acyclic (correctly skips to avoid unbounded longest-path under a cycle). Boundary is exactly right: 4-node/3-edge chain = depth 3 → passes; 4-edge chain → over-depth. Matches the pinned edges-definition (GV cond 2).
- **Combined DAG** (GV cond 3) honoured — cycle across mixed instance_of/subclass_of/part_of edge types is caught; part_of-misuse correctly left as the separate Q2 soft-flag, not a fuzzy hard-reject.
- **Attribution diff** (with-candidate vs without) is a nice touch — a pre-existing store violation doesn't reject an unrelated import; only candidate-introduced violations fail. Correct.

**Both arms verified** — tests assert each check rejects the bad case AND passes the good case (malformed/well-formed, missing/existing, cycle/DAG, depth-4/depth-3). Guard Testability (t/2971) satisfied: pure checker, no IO.

**GV conditions all met:** depth-def + combined-DAG pinned in comments at point-of-use (Co-Location ✓); cond-4 gate-clean on the current store (0 edges — honest: relations aren't persisted yet, so the gate binds when population starts). Validate-only semantics documented so it's not misread as a no-op.

**Deviations accepted** (all flagged at t/3170#3): Q3 → `-Report relation-dag` kind on Get-EntityReport (the home I suggested); Q4 interim → well-formed term:* accepted-without-existence until Shared Lib parity lands (ent-* existence IS enforced; well-formedness+cycle+depth still bind on term edges) — correct interim, tracked.

**One non-blocking observation** (add a comment, no code change needed): depth is skipped for the WHOLE graph if any cycle exists, so a candidate introducing both a cycle (one component) and an over-depth path (another) reports only the cycle. Fine for a reject-gate — it rejects either way and over-depth surfaces on re-run after the cycle's fixed — and under the clean-store invariant (cond-4 + Get-EntityReport) a pre-existing cycle masking a candidate over-depth can't arise. Worth a one-line comment noting the interaction; not a merge gate.

Verified: base=main ✓, head d8bd5289. Un-draft + self-merge once test-powershell is green on this OID. Excellent, careful gate — the attribution diff and the StrictMode-hardened iterative traversals are exactly the right level of rigor.

PR #1733, opened 2026-09-01
feat(devops): paid-fallback overflow cost alert (t/3110)

jpsnover:
Lands t/3110 — cost-alert on the PAID Gemini fallback key. Unblocked by t/3143 (GEMINI_PAID_KEY now live in prod).

## What
New Azure Monitor scheduled-query alert **`alert-paid-fallback-overflow`** in `deploy/azure/main.bicep`:
- **Signal:** `ContainerAppConsoleLogs_CL` where `Log_s contains "Paid fallback succeeded"` — the log emitted at `routes/ai.ts:187` when `generateWithPaidFallback` completes a **billed** call.
- **Threshold:** >0 occurrences in a rolling **60 min** (`windowSize PT1H`, `evaluationFrequency PT15M`) — distinguishes sustained overflow from a lone transient.
- **Severity 2 (Warning)**, routed to the existing `ag-aitriad-restart-alert` action group (emails `ALERT_EMAIL`), same pattern as the other 7 Bicep alerts.
- Runbook alert table updated (`production-release.md`).

## Why
Per the free-primary/paid-overflow design (t/3111), the default anon path is free (4-key pool, 30 RPM front-door), so paid-fallback usage should be **~0**. Any sustained nonzero = the free pool is under-provisioned or throttling into billable overflow — a real operational signal, not steady-state cost. Gives us the data to decide if a 3rd/4th free key is ever needed (per TL p/331#800).

## Verification
- `az bicep build` compiles clean (exit 0); `alert-paid-fallback-overflow` present in the compiled ARM JSON.
- No GCP dependency — the signal is entirely Azure-side (supersedes the original "GCP budget alert" framing in the ticket).

## Gate Verification (draft-held for TL, per t/3111 cost/limits surface)
Both arms, provable now that the paid key is live:
- **FIRE:** with `GEMINI_PAID_KEY` set, force free-pool exhaustion → `Paid fallback succeeded` logs → alert fires.
- **CLEAN:** default free path → no such log → silent.

Held as **draft** pending TL GV sign-off (cost-governance per t/3111). Un-draft on approval.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1731, opened 2026-09-01
fix(server/ai): grounded-search key rotation + retry + paid-overflow parity (t/3175)

jpsnover:
Fixes the debate search 429 (t/3165 sub-issue). Both TL decisions approved (p/522#117/#118); design in t/3175#2. → **Main (TL) GV** (AI-resilience surface).

## Root cause
`generateTextWithSearch`'s Gemini path used only `explicitApiKey[0]` and called `geminiGroundedSearch` **directly** — no rotation, no retry, no paid-fallback. A debate round's search-verify bursts hammered **one** free-tier key while the other 3 sat idle → `429 api_key_exhausted`, no recovery. `generateText` has the full stack; search never did (the keyRotator audit claim was stale).

## Fix (full parity with generateText)
1. **LOAD-BEARING** (`aiBackends.ts`): wrap the grounded-search call in `withRetry(callWithKeyRotation('gemini', keys, apiKey => geminiGroundedSearch(...)))`. Distributes across the whole pool + backs off. Fixes it **with the existing keys** (pool was under-utilized, not exhausted).
2. **PARITY add-on** (`routes/ai.ts`): `generateWithSearch` gains `generateWithPaidFallback`-style paid overflow — free-tier 429 + pool exhausted → 3s throttle → retry once with the paid key (never in the free rotation). **No-op until `GEMINI_PAID_KEY` is set (t/3143)**; ships independently of it.

## Tests
`searchKeyRotation.test.ts` (3/3): 429-on-first-key **rotates to a different key + recovers**; whole-pool-429 **surfaces the 429** for outer backoff (tried each key once); healthy search returns text+citations unchanged. tsc clean.

The paid-overflow wrapper structurally mirrors the already-tested `generateWithPaidFallback` (`paidFallbackOrdering.test.ts`) — flag if you want a dedicated route-level test at GV.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1735, opened 2026-09-01
fix(server/ai): search paid-success emits shared 'Paid fallback succeeded' marker (t/3110)

jpsnover:
Follow-up to #1731 (TL, p/522#125). DevOps's paid-overflow cost alert (#1733, t/3110) counts free→paid overflow by matching the log token **"Paid fallback succeeded"**. The non-search `generateWithPaidFallback` emits it; #1731's search paid-success logged `generate+search success …(paid fallback)` — a different string, so the alert would **miss debate-search overflow** (the t/3165 second front).

**Fix:** the search paid-**success** record now carries the shared marker. The attempt log and failure log keep their own wording, so only billed successes are counted (no double-count / false positives).

**Test:** `searchPaidOverflow.test.ts` asserts the paid-success record carries the marker AND the attempt record does not. 5/5.

**Self-cert /trivial-change** — one message string + a test assertion, single scope, no behavior/API change.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/searchPaidOverflow.test.ts
+++ b/taxonomy-editor/src/server/__tests__/searchPaidOverflow.test.ts
@@ -112,5 +112,13 @@ describe('generateWithSearch — paid-overflow arm (t/3175)', () => {
     const fired = frRecords.find(r => r.type === 'ai.fallback');
     expect(fired).toBeDefined();
     expect((fired!.data as Record<string, unknown>).fallback).toBe('paid');
+    // t/3110: the paid-SUCCESS record must carry the shared "Paid fallback succeeded" marker
+    // DevOps's overflow-cost alert (#1733) keys on — same token as generateWithPaidFallback.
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL cross-consistency review — APPROVED (t/3110 / t/3165 second front).

Verified the alert↔emit token contract across #1735 (this) and #1733 (DevOps alert):
- #1733 matches `Log_s contains "Paid fallback succeeded"`; this PR's search paid-success emits `Paid fallback succeeded (search) …` → **contains the exact contract substring**, so the alert fires on debate-search overflow (closes the miss you identified).
- Both-arms test is correct: paid-**success** carries the marker (counted) AND the **attempt** log does NOT (`not.toContain`) → no double-count / false positive.

Signal integrity confirmed. Merge on green + pre-self-merge verify. No strict order vs #1733 — land both for full search+non-search overflow coverage (each is benign without the other: alert without emit just doesn't count search; emit without alert is an unwatched but harmless token).

PR #1748, opened 2026-09-01
feat(embeddings): off-thread worker embedding module (t/3181, item A) [GATED: TL GV]

jpsnover:
## t/3181 — off-thread worker embedding module (item A) · **GATED on Main TL Gate Verification**

The shared prerequisite for the durable t/3165 fix; unblocks B (t/3183 web) + C (t/3184 Electron). Moves ONNX embedding compute onto a `worker_thread` so a large batch can't starve the main event loop. Design GV-approved (t/3181#2). **Draft until GV clears.**

### What ships
- **`embeddingWorker.ts`** (new) — worker entry; owns the ONNX session exclusively, pins intra/interOp=1 (C2, worker-only), per-chunk heartbeat, packed fp32 vectors via a **transferred** ArrayBuffer.
- **`offThreadEmbedding.ts`** (new) — main-thread manager + the contract: `computeEmbeddingsOffThread(texts, opts?: {requester?})`. Single worker, bounded FIFO (16), heartbeat watchdog **armed on dispatch** (8000ms, reset per heartbeat), respawn-backoff (250ms→5s). Every shed/crash/wedge fails **loud** (ActionableError → t/3078 503) with a **requester-named WARN**. **No in-thread fallback** anywhere. Tunables co-located with trade-off comments (Gate Co-Location).
- **`onnxEmbedding.ts`** (changed) — Q3 **worker-only**: session thread-options parameterized (`setSessionThreadOptions`); in-thread `create()` is **byte-identical** to the pre-t/3181 baseline. Optional `onChunkDone` heartbeat callback + exported `EMBEDDING_DIM`.

### Conditions (t/3181#2)
- **4-arm proof** (`offThreadEmbedding.test.ts`, injected fake worker): (a) queue-saturation → shed + WARN(requester, depth); (b) crash → reject + respawn-backoff; (c) wedge → terminate + respawn + WARN; (d) worker failure → **throws ActionableError, never computes in-thread**. + a marshaling round-trip.
- **Real-ONNX equivalence** — skip-guarded with a **visible skip** (no CI job runs onnxruntime-node + model; `python-embed-smoke` is the Python stack). Follow-up **t/3198** wires it into CI when the model lands there.
- **Condition 3 (verified locally, real ONNX CPU):** intra/interOp=1 vs default is **BIT-EXACT** — `maxAbsDiff=0, mismatches=0/1536` across 4 texts (incl. a long multi-chunk one) × 384 dims. No tolerance needed.

### Verify
tsc main+server **0** · eslint **0** · vitest **6 pass / 1 skip** · existing `onnxEmbedding` suite still green.

Contract locked for B/C. PR → **Main TL** for GV with the four-arm proof + reported equivalence finding.

### DIFF
--- a/lib/embeddings/embeddingWorker.ts
+++ b/lib/embeddings/embeddingWorker.ts
@@ -1,0 +1,76 @@
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+/**
+ * Off-thread embedding worker entry (t/3181, item A — the shared prerequisite for B/C).
+ *
+ * Runs in a Node/Electron `worker_thread`, owns the ONNX session EXCLUSIVELY, and computes
+ * embeddings off the main event loop so a large batch can't starve request handling (the durable
+ * fix for the t/3165 starvation class). The manager (`offThreadEmbedding.ts`) is the only caller.
+ *
+ * Protocol (mirrored by the manager):
+ *   in :  { id, texts }
+ *   out:  { type:'heartbeat', id, chunk }                              — forward-progress ping
+ *         { type:'result', id, ok:true, buffer, count, dim }          — packed vectors, buffer TRANSFERRED
+ *         { type:'result', id, ok:false, error }                      — compute failed (manager load-sheds)
+ *
+ * Design invariants (t/2977#6):
+ *   - SINGLE worker-owned session (this module is the only place ONNX is created post-migration —
+ *     the manager never warms/computes in-thread, so the ~250MB model is resident once, not twice).
+ *   - ONNX intra/interOp pinned to 1 (C2) so onnxruntime can't oversubscribe the host's 1–2 vCPUs.
+ */
+
+import { parentPort } from 'node:worker_threads';
+import { computeEmbeddings, setSessionThreadOptions, EMBEDDING_DIM } from './onnxEmbedding.js';
+
+interface EmbedRequest {
+  id: number;
+  texts: string[];
+}
+
+if (!parentPort) {
+  throw new Error('embeddingWorker.ts must be run as a worker_thread — parentPort is null');
+}
+const port = parentPort;
+
+// C2 (worker-ONLY, TL Q3): pin ONNX to a single op thread BEFORE the session is created, so the
+// worker's inference can't spin up a thread per core and contend with the main event loop. The
+// in-thread path deliberately keeps its default threading (byte-identical baseline) — this pin
+// lives here, not module-wide.
+setSessionThreadOptions({ intraOpNumThreads: 1, interOpNumThreads: 1 });
+
+async function handleRequest(msg: EmbedRequest): Promise<void> {
+  const { id, texts } = msg;
+  try {
+    // Emit a forward-progress heartbeat after each chunk resolves. The manager's watchdog resets on
+    // each one, so a healthy-but-slow large batch is not false-killed, while a genuinely wedged
+    // worker (no chunk completes) is caught within ~one chunk's timeout window.
+    const vectors = await computeEmbeddings(texts, (chunk) => {
+      port.postMessage({ type: 'heartbeat', id, chunk });
+    });
+
+    // Pack N×dim fp32 into ONE contiguous Float32Array and TRANSFER its ArrayBuffer (zero-copy — the
+    // manager slices it back into N views). Empty batch → zero-length buffer, still a valid transfer.
+    const dim = vectors.length > 0 ? vectors[0].length : EMBEDDING_DIM;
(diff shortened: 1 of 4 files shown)


jpsnover:
## TL Gate Verification — 1 REQUIRED fix (crash-recovery race), then APPROVED. Excellent implementation.

I read all three impl files end-to-end. Binding requirements (t/2977#6 / t/3181#2) are all correctly realized:
- **C1 single worker/session** ✓ — module-level singleton, no pool; session lives only in the worker (main never inits → no double ~250MB footprint).
- **C2 intra/interOp=1 WORKER-ONLY** ✓ — `setSessionThreadOptions({...:1})` in the worker before create; in-thread `create()` reduces to `{executionProviders}` byte-identical (Q3 exactly right — the parameterize-not-module-wide approach preserves the flag-off equivalence baseline).
- **C3 watchdog armed ON DISPATCH** ✓ (my Q1 refinement — catches a pre-first-heartbeat/model-load hang), reset per chunk-heartbeat; respawn-with-backoff (250ms→5s, reset on clean task); shed during the gap.
- **Bounded queue (16) → shed + WARN with requester** ✓; **no in-thread fallback** ✓ — every failure path is an ActionableError → server load-shed, and the code documents *why* (would reintroduce t/3165). This is the load-bearing invariant and it's airtight.
- **Condition-3 (bit-exact)** ✓ — your maxAbsDiff=0, 0/1536 confirms the reduction-order reasoning (intra-op parallelizes across output elements, not within a dot-product). Visible skip-guard + t/3198 follow-up honored.

### REQUIRED FIX — stale-worker events can tear down the *live* worker (crash-recovery race)
`spawnWorker()` attaches `message`/`error`/`exit` listeners that reference module globals, but they are **never detached**, and the handlers don't check whether the emitting worker is still `_worker`. Sequence:
1. W1 wedges → `onWorkerDown('wedge')`: `_down=true`, `safeTerminate(W1)` (async), `_worker=null`, respawn timer = **250ms** (first backoff).
2. Backoff clears at +250ms → `pump()` spawns **W2**, `_worker=W2`.
3. W1's `terminate()` completes and emits **`exit` (code≠0)** — if this lands **after** the 250ms gap (terminate lagging under load — the exact saturation condition we're in), W1's exit listener re-enters `onWorkerDown('exit')` with `_down=false` → **`safeTerminate(_worker)` kills W2** and rejects W2's in-flight task → spurious cascading respawn.

The `if (_down) return` guard only covers the case where the stale exit lands *during* the gap; a terminate slower than the 250ms first backoff escapes it. Narrow, but it's in the reliability-critical path of the incident's durable fix, and when it hits it makes the manager thrash precisely under stress.

**Fix (small):** bind each handler to its worker instance and ignore events from a non-current worker — capture `const w` in `spawnWorker()` and guard `if (w !== _worker) return;` at the top of the message/error/exit handlers (generation/identity guard). This makes stale-worker events inert regardless of terminate timing, and subsumes the stale-result guard more robustly. **Add a test:** fake worker emits `exit` *after* a respawn → assert the new worker survives and its in-flight task is unaffected.

### Everything else: approved
The injectable-factory testability, the ActionableError discipline, the Gate Co-Location on all four constants, and the transfer/unpack zero-copy path are all exactly right. Re-request GV after the identity-guard fix + its test; then un-draft → I clear it. This is the largest/riskiest change in the incident and it's very cleanly done — one race is a good outcome for a worker-lifecycle manager.

jpsnover:
## TL Re-GV — CLEARED. Un-draft + self-merge on green (head 7530f055).

Verified the required fix directly:
- **Identity-guard on ALL THREE handlers** ✓ — `message`/`error`/`exit` each check `w === _worker` (captured per-instance in `spawnWorker`), so a superseded worker's late event is dropped instead of acting on current state. Correct + clearly commented.
- **Test genuinely exercises the race** ✓ — worker1 crashes → respawn-backoff cleared → worker2 serving; then the OLD worker fires late `exit(137)` + a stale `message(staleId)`, and it asserts: worker2 **not** terminated, **no** crash/wedge WARN from the stale events, and task 2 resolves normally. This discriminates — under the pre-fix code the stale `exit(137)` would have re-entered `onWorkerDown` and torn down worker2. Proper regression test for the exact failure mode.
- CI fully green (25/25 incl. test-electron running the 8-test suite, test-powershell, ci-gate).

All binding requirements + the crash-recovery race now verified. **This is the critical-path prerequisite for the entire t/3165 durable fix — B (t/3183) + C (t/3184) unblock the moment this lands.** Base=main, head=7530f055 matches your push. Un-draft → self-merge. Excellent work, and fast — one race caught + cleanly fixed on the riskiest change in the incident. Ping me when it's on main and I'll confirm B/C are unblocked.

PR #1750, opened 2026-09-01
fix(embeddings): identity-guard worker handlers (t/3181 — recover stranded GV fix)

jpsnover:
Lands the t/3181 identity-guard fix that was **stranded**: #1748 merged at the pre-fix head a77ed078 (merge 594a503a), so main got the buggy version and the fix commit 7530f055 never landed. This is that exact 2-file diff cherry-picked onto current main.

**The fix (TL GV-cleared, p/342#234):** capture the worker ref in `spawnWorker` and identity-guard all three handlers (`w === _worker`) so a superseded/terminated worker's late error/exit/message can't tear down the healthy respawned worker. Plus the respawn-survives test (old worker's late exit(137)+stale message after respawn → new worker untouched, no spurious WARN, task completes; genuinely discriminates vs pre-fix).

Identical content already verified: vitest 7 pass/1 skip, eslint 0, CI 25/25 on 7530f055. Cherry-pick applied clean onto main (main already had a77ed078's base files via #1748). Self-merge on green — GV already granted.

Unblocks B (t/3183) + C (t/3184) once on main.

### DIFF
--- a/lib/embeddings/offThreadEmbedding.test.ts
+++ b/lib/embeddings/offThreadEmbedding.test.ts
@@ -126,6 +126,39 @@ describe('offThreadEmbedding — arm (b): crash → reject + respawn with backof
   });
 });
 
+describe('offThreadEmbedding — respawn survives a stale event from the superseded worker (identity-guard)', () => {
+  it('ignores a late exit/message from the old terminated worker; the fresh worker keeps serving', async () => {
+    vi.useFakeTimers();
+    const workers: FakeWorker[] = [];
+    __setEmbeddingWorkerFactory(() => { const w = new FakeWorker(); workers.push(w); return w; });
+
+    // Task 1 dispatched on worker1, then worker1 crashes → reject + respawn-backoff.
+    const p1 = computeEmbeddingsOffThread(['x'], { requester: 'first' });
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL confirm — stranded-fix recovery verified. Cleared to self-merge.

This correctly recovers the #710 stale-head stranding: #1748 merged at pre-fix head `a77ed078`, so main shipped the buggy worker module and the GV-cleared fix (`7530f055`) never landed. #1750 is that exact 2-file identity-guard diff cherry-picked onto current main.

- **GV already granted** (p/342#234) — the identity-guard on all three handlers (`w === _worker`) + the respawn-survives test (old worker's late exit(137)/stale message can't tear down the healthy respawn). Directly satisfies SO condition **C3** (wedge/respawn resilience).
- **Content matches** the GV'd 7530f055 (vitest 7/1-skip, eslint 0); CI green on #1750.
- **Sequencing OK:** no B (ServerAPI) / C (ElectronMain) delegation PRs open yet → this lands first, so B/C build against the corrected worker module, not the stranded one.

Self-merge on green (pre-self-merge: base=main, head-match, CI-green-on-OID). Good catch on the stranding — this is the exact class the pre-self-merge verify exists to prevent.

PR #1753, opened 2026-09-01
perf(embeddings): off-thread ONNX compute behind EMBEDDING_WORKER_OFFLOAD (t/3183)

jpsnover:
## t/3183 — Item B of t/2977 (the t/3165 durable fix, web arm)

Consumes the shared off-thread embedding worker (**t/3181 / #1748, now on main**). `computeEmbeddings` routes its ONNX miss-text pass to `computeEmbeddingsOffThread` behind the per-build env switch **`EMBEDDING_WORKER_OFFLOAD`** (default **OFF**, web-first). Flag OFF is **byte-identical** to today's in-thread path; flag ON keeps the event loop responsive under a large novel-text batch (the t/3165 starvation). Cache resolve + chunk-and-yield stay on the main thread unchanged.

**Ships flag-OFF → zero prod behavior change** until the deliberate post-GV canary flip.

### Changes
- **`config.ts`** — `isEmbeddingWorkerOffloadEnabled()`: env, per-call, default-OFF on unset/blank/non-true (case-insensitive `1`/`true`), documented at point-of-use.
- **`ai/aiBackends.ts`** — the `onnx-batch` chain member routes to the worker when enabled (`Float32Array[]`→`number[][]`); `requester` threaded through so a worker shed WARN names the caller; **no in-thread fallback** on shed (fail-loud → 503, matching the worker's contract). C6 demand cap-and-WARN `NOVEL_TEXT_DEMAND_BASELINE=256`.
- **`routes/ai.ts`** — `/api/embeddings/compute` passes `requester='embeddings-compute'`; tiering comment vs `LARGE_RECOMPUTE_WARN_ITEMS=64`.
- **`__tests__/embeddingWorkerOffload.test.ts`** — 5 tests.

### TL GV context (Q1/Q2 absorbed — t/3183#3)
- **Q1:** capability switch = **env var** (per-revision deploy toggle; flag-off = instant rollback; consistent with sibling `EMBEDDING_COMPUTE_CHUNK`). Not `feature-flags.json`.
- **Q2:** `256` is **PROVISIONAL + WARN-only** (never a hard reject), **tiered** vs `LARGE_RECOMPUTE_WARN_ITEMS=64` (64 = "notable volume"; 256 = "demand itself may be the bug"). Calibrated from the storm-replay canary's real novel-per-turn (cacheHits=0) distribution **before flag-on** — tracked as **t/3199** (blocks the flip). Not confirmed by fiat (t/3085).

### Verification
- **Both-arms liveness proof** (the assertion the incident lacked): a 1347-text compute keeps the main-thread max event-loop gap **under** the deadline with the flag ON, and **starves past** it with the flag OFF.
- 4 contract tests: flag-OFF byte-identical, flag-ON worker routing + `Float32Array→number[]` widen, `requester` threaded, cap-and-WARN above/at baseline.
- `embeddingWorkerOffload.test.ts` **5/5**; 28/28 regression across embeddings/search suites; `tsc -p tsconfig.server.json` clean on touched files.

→ **Main (TL) GV.** Un-draft + self-merge on approval (pre-self-merge verification). Then line up the storm-replay canary (t/3199).

### DIFF
--- a/taxonomy-editor/src/server/__tests__/embeddingWorkerOffload.test.ts
+++ b/taxonomy-editor/src/server/__tests__/embeddingWorkerOffload.test.ts
@@ -1,0 +1,158 @@
+// @vitest-environment node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// t/3183 (t/2977 Item B) — computeEmbeddings delegates the ONNX miss-text pass to the shared
+// off-thread worker (t/3181) behind the per-build capability switch EMBEDDING_WORKER_OFFLOAD.
+//
+// The headline proof is the BOTH-ARMS LIVENESS test the t/3165 incident lacked: under a ≥1347-text
+// compute, the main-thread event-loop max-gap stays under the ACA liveness deadline with the flag
+// ON (work is off-thread) and BLOWS the deadline with it OFF (a 128-text ONNX chunk blocks the loop
+// synchronously). The other four tests lock the contract: flag-OFF byte-identical, flag-ON routes to
+// the worker + widens Float32Array→number[], the requester label reaches the worker, and the C6
+// demand-baseline WARN fires above the by-design novel-text ceiling / stays silent at it.
+//
+// The REAL resolveEmbeddings + resolveEmbeddingsChunked run (chunk-and-yield is the mechanism under
+// test); only the two leaf computes (onnx in-thread / off-thread worker) and the data-file read are
+// mocked. isPythonEmbeddingAvailable is forced false so the chain is onnx-only (the ACA shape).
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { performance } from 'node:perf_hooks';
+
+const { onnxCompute, offthread, records } = vi.hoisted(() => ({
+  onnxCompute: vi.fn(),
+  offthread: vi.fn(),
+  records: [] as Array<{ level?: string; message?: string; data?: Record<string, unknown> }>,
+}));
+
+// onnx-only chain (Python venv absent = the ACA shape). tryWarmup true → the onnx member is pushed.
+vi.mock('../../../../lib/embeddings/onnxEmbedding.js', () => ({
+  tryWarmup: vi.fn(async () => true),
+  warmup: vi.fn(async () => true),
+  computeEmbedding: vi.fn(async () => []),
+  computeEmbeddings: onnxCompute,
+}));
+// The worker contract (t/3181) — mocked so this test never spawns a real thread.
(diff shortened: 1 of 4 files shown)


jpsnover:
## TL GV — APPROVED. Un-draft + self-merge on your pre-self-merge check.

Faithful implementation of the approved design (t/3181#2 contract) with both Q1/Q2 answers absorbed. Verified end-to-end:

- **Flag default-OFF** ✓ — `isEmbeddingWorkerOffloadEnabled()` true only for `1`/`true`, unset→false; env var (not feature-flags.json — correct, it's a build/deploy capability + ops canary flip, matches sibling EMBEDDING_COMPUTE_CHUNK); read per-call so the liveness test can toggle it.
- **Miss-only routing** ✓ — delegation is the chain's `compute` member (`enabled ? computeEmbeddingsOffThread(t,{requester}).then(widen) : onnxComputeEmbeddings(t)`), so only miss-texts marshal to the worker; cache RESOLVE stays main-thread (the HLD boundary). Float32Array→number[][] widen present.
- **No in-thread fallback** ✓ — a worker shed/crash rejects → load-shed 503, explicitly NOT caught + recomputed in-thread (would reintroduce t/3165). The comment states it; the worker already WARNs the shed with requester.
- **C6 demand cap-WARN** ✓ — `cacheMisses > 256` → one WARN (requester + cacheMisses + baseline + `provisional:true`), never a hard reject; **tiered** vs LARGE_RECOMPUTE_WARN_ITEMS=64 documented at BOTH point-of-use; 256 flagged PROVISIONAL with the calibrate-at-canary rationale (t/3199 gates flag-on). Exactly the p/522#133/#134 guidance.
- **requester threaded** ✓ (routes/ai.ts → 'embeddings-compute').
- **Both-arms liveness** ✓ — the assertion the incident lacked: flag-OFF reproduces the starvation (max loop-gap ≥ deadline), flag-ON stays live (< deadline, worker called, in-thread never touched).

**Two notes, neither blocking:**
1. The unit liveness test correctly MODELS the worker as non-blocking (setImmediate resolve) — right for this layer (it proves the *delegation* keeps the server loop live). The REAL-env liveness proof (real ONNX in a real worker under real load) is the **storm-replay canary** — already the close gate. No gap, just correct layering.
2. The demand-WARN fires on BOTH arms (flag-off too) — that's intended + good: it makes unbounded demand visible NOW, before the offload is even enabled. It's pure observability (a log record), so flag-off compute output/path stays byte-identical.

**Flag stays OFF on merge** — the canary calibrates the 256 baseline + proves real-env liveness (t/3199) before flip-on. Un-draft + self-merge on your pre-self-merge verify (base=main, head==ebf550aa, CI green on that OID). Then C (t/3184) + D (t/3182) + the canary. Excellent work — this is the web path of the durable fix, cleanly done.

PR #1757, opened 2026-09-01
feat(devops): scheduled Gemini key-pool health check (t/3142)

jpsnover:
Implements t/3142 per TL-approved design (t/3142#3). New workflow `key-pool-health.yml`.

## Mechanism
Daily (06:17 UTC) + `workflow_dispatch`. Validates the **deployed** `FREE_TIER_GEMINI_KEY` pool via t/3141's `Test-GeminiKeyPool` (auth-only probe, **0 tokens**, masked fingerprints only — never a raw key). Secret written to a `$RUNNER_TEMP` file, deleted in a finally block.

## Alert logic (TL-approved, two-rung)
- **DEAD>0** (any 401/403) → page (early-warning: permanent capacity loss, replace before we near the floor).
- **usable (Valid+Throttled) < FLOOR(2)** → page (critical depletion; 2 = the locked V=12/K=2 front-door minimum).
- **THROTTLED = healthy → 429 NEVER pages** (load-bearing correctness).
- **Confirm-before-page:** one auth-only re-probe on any would-be alert / network-err; second result authoritative (kills flaky pages).
- **OBSERVE-ONLY** (`PAGE_ENABLED=false`): logs would-be alerts, opens NO issue, until TL flips to paging after ≥1 green real-env cycle (Gate Promotion).
- Paging (when enabled) opens/refreshes a `key-pool-alert` issue with the masked table + @owner, auto-closes on recovery (health-monitor pattern).

## Both-arms evidence (offline)
- **Logic (exact workflow expression), 6/6:** healthy-4 → silent · pure-throttle(T4) → **silent (429-not-fail)** · V2/T2 → silent · one-DEAD → **fire** · usable<2 → **fire** · network-err → **fire**.
- **Real integration (Test-GeminiKeyPool live probe):** a deliberately-bad `AQ.` key → `DEAD(401)` → usable=0 → `alert=True`, masked `AQ.Ab8…dXYZ`.

## Gate Co-Location
Floor(2)+its tie to V=12/K=2, throttled=healthy, and the observe→paging promotion rule are all documented inline in the workflow header.

## Draft — held for TL both-arms GV + the promotion cycle
Per t/3142#3: route to TL GV; after sign-off, run observe-only ≥1 green real-env cycle against the live pool, then TL flips `PAGE_ENABLED=true`. No deviation from approved design.

### DIFF
--- a/.github/workflows/key-pool-health.yml
+++ b/.github/workflows/key-pool-health.yml
@@ -1,0 +1,192 @@
+# ─── Free-tier Gemini Key-Pool Health ───
+# What:  Daily validates the DEPLOYED free-tier Gemini key pool (the
+#        FREE_TIER_GEMINI_KEY secret — the authoritative live pool; GitHub
+#        secrets are write-only, so a job holding the secret is the only durable
+#        way to know per-key validity). Pages when the pool degrades. (t/3142)
+# Why:   Free `AQ.`-format keys can expire/revoke SILENTLY — today nothing
+#        notices until generations start 401'ing mid-debate (t/3165 secondary).
+# How:   Reuses Test-GeminiKeyPool (t/3141) — auth-only probe per key
+#        (GET /v1beta/models?key= — ZERO tokens), masked fingerprints only,
+#        NEVER prints a raw key.
+#
+# ── Gate config (co-located — change deliberately) ──
+#   FLOOR = 2  → page when usable (Valid+Throttled) < 2. 2 is the LOCKED
+#     minimum-viable pool: front-door capacity = min(V×K, 30) with V=12
+#     (FREE_TIER_RPM_PER_KEY, proxyTiers.ts) and K = pool size; K=2 gives
+#     min(12×2,30)=24 ≥ 1.2·T (debate burst T≈20). Below K=2 the front door
+#     collapses → critical. If V or the T target changes, re-derive this floor.
+#   THROTTLED = HEALTHY  → a 429 key is rate-limited, not dead; it self-heals
+#     (rotation + paid overflow absorb bursts). Paging on 429 = noise, so
+#     Throttled counts toward "usable" and NEVER pages. (The load-bearing arm.)
+#   DEAD>0 (early-warning rung) → any 401/403 key is permanent capacity loss;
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL both-arms GV — APPROVED (merge as observe-only). Promotion to paging is the separate evidence-gated step.

Faithful to the approved design (t/3142#3); verified every arm in the workflow:
- **Throttled = HEALTHY (the load-bearing arm)** ✓ — `usable = Valid + Throttled`, and 429 never contributes to `alert`. A pure-throttle pool does NOT page. This is the arm that would've made it noise; it's correct.
- **alert = Dead>0 OR usable<floor OR NetworkErr>0** ✓ — the two-rung design (any DEAD → early-warning page; usable<2 → critical) as specified.
- **Confirm-before-page** ✓ — first trip → 20s sleep + re-probe, second result authoritative; kills flaky pages.
- **Observe-only** ✓ — `PAGE_ENABLED='false'` default; the page step is gated `== 'true'`, and a distinct step logs the would-be alert without opening an issue. Correct Gate-Promotion posture.
- **Gate Co-Location** ✓ — FLOOR=2 tied to the locked V=12/K=2 minimum (+ re-derive-if-V/T-changes), throttled-semantics, and the promotion threshold all documented in the header at point-of-use. Exactly right.
- **Secret handling** ✓ — secret env-scoped to the step, written to $RUNNER_TEMP, deleted in `finally`; auth-only probe, masked fingerprints, never prints a raw key. Pinned action SHAs; auto-close on recovery; absent-secret treated as degraded.

**One watch-item for the observe cycle (not blocking):** the `NetworkErr>0` arm — a persistent Gemini-endpoint blip would page even though it's not a key problem. The re-probe filters transients, so it should be quiet; the observe-only cycle is exactly where we confirm NetworkErr isn't a false-page source before promotion.

**Promotion path (evidence-gated, per Gate Promotion):** merge observe-only → `workflow_dispatch` ≥1 run against the live 4-key pool → confirm it reads healthy (no false alert, NetworkErr quiet) → THEN I flip `PAGE_ENABLED='true'` via the one-line promotion PR. Un-draft + self-merge (observe-only) on green. Clean gate — good design-first discipline.

PR #1758, opened 2026-09-01
test(embeddings): both-arms unit test for the large-recompute volume WARN (t/3188)

jpsnover:
## t/3188 — #1720 GV-condition close-out

#1720 shipped the `LARGE_RECOMPUTE_WARN_ITEMS=64` volume WARN in `routes/ai.ts` **without** the both-arms test the GV required (merge landed before the condition was met). This adds it, so the threshold predicate can't silently rot.

**Both arms** (route-handler harness mirroring `embeddingsBatchCapAnd503`, with shared recorder + log spies so the WARN is assertable on both sinks):
- **FIRES:** `texts.length >= 64 && cacheHits === 0` → volume WARN on recorder + api log.
- **SILENT:** 63 texts (below threshold); and a large batch (100) that resolves from cache (`cacheHits > 0`).

Pure **/add-test** addition (self-cert) — the WARN is observability, no API/behavior change. 3/3 green locally.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/embeddingsLargeRecomputeWarn.test.ts
+++ b/taxonomy-editor/src/server/__tests__/embeddingsLargeRecomputeWarn.test.ts
@@ -1,0 +1,199 @@
+// @vitest-environment node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// t/3188 — both-arms unit test for the large-recompute VOLUME WARN in POST /api/embeddings/compute
+// (routes/ai.ts). #1720 shipped `LARGE_RECOMPUTE_WARN_ITEMS=64` WITHOUT the both-arms test the GV
+// required, so the threshold predicate was untested (silent rot risk). This closes that gap:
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL — GV-condition MET (closes the deferred both-arms from #1720).

Verified the both-arms genuinely discriminates the volume WARN:
- **FIRES:** `texts.length >= 64 && cacheHits === 0` — volume-of-uncacheable, the exact t/3165 incident fingerprint (large novel-text recompute).
- **SILENT:** 63 texts (below threshold) AND — the load-bearing negative — a **100-text batch that resolves from cache (cacheHits>0)**. That second case is what makes the predicate correct: the WARN targets uncacheable *volume*, not large batches that legitimately resolve. Without it, the WARN would false-fire on every large corpus embed.

Asserted on both sinks (recorder + api log). `/add-test` self-cert accepted — pure observability, no behavior change.

Merge on green (`test-electron (taxonomy-editor)` pending — it runs this test) + pre-self-merge verify. This closes the observability arm I asked for at t/3165#25 (make the silent large-no-ids recompute greppable) and the GV-condition #1720 shipped without.

PR #1768, opened 2026-09-01
feat(grounding): G8a inline write-hook — debounced scoped reconcile after PUT /api/taxonomy/:pov (t/3171)

jpsnover:
## t/3171 (G8a) — inline grounding write-hook → Main (TL) GV

Built to TL's design spec (t/3171#2), on CL's scoped reconciler entry point (#1736). After a `PUT /api/taxonomy/:pov` write succeeds, the changed node ids are enqueued for a **debounced, fire-and-forget** scoped grounding reconcile — behind the **`grounding_reconcile_inline` feature flag (default OFF)**.

### ⚠️ Enable sequencing (hard, stated per TL spec)
Ships **flag-OFF**. Do NOT enable until BOTH land/confirmed: the reconciler tool-lock (**t/3194**, landed — `grounding_lock` is in `reconcile_grounding.py`) AND the PS cmdlet lock-honoring (**t/3203**, open). Enabling before then lets a concurrent unlocked writer lost-update the shared `entity_mentions.json`. Flag-OFF = zero behavior change (the route adds only a `getFlag()`).

### Changes
- **`groundingReconcileHook.ts`** (new): in-process dirty-set + debounce (`DEBOUNCE_QUIET_MS=3000` / `DEBOUNCE_MAX_WAIT_MS=15000`, co-located); single-child-at-a-time in-process guard (belt-and-suspenders over the reconciler's own cross-process lock); `execFile` the scoped `--nodes <ids> --apply` reusing the server's `PYTHON` pattern; parse the leading stats JSON for the FR `changed` field. **Error isolation:** non-zero/timeout/spawn → `ActionableError` + FR **WARN**, never propagated (a failed grounding refresh must not fail the taxonomy write — the #1737 lesson). Node ids are **allowlist-sanitized** before the subprocess (`sanitizeNodeIds`).
- **`routes/taxonomy.ts`**: flag-gated `diffNodes(old, incoming)` → `enqueueGroundingReconcile` **after** the 200.
- **`config.ts`**: `RECONCILE_SCRIPT` off `PROJECT_ROOT`.

### Tests (12; TL's both-arms plan)
Hook unit (fake timers + injected runner — no python/data-repo spawned): FIRES / UNCHANGED-no-op / COALESCE-burst-into-one / MAX_WAIT-freshness / ERROR-ISOLATION / SINGLE-CHILD-guard / sanitize / parseStats. Route-wiring: flag ON changed→enqueue, ON new→enqueue, ON unchanged→no-enqueue, OFF→no-enqueue. `routeTable` regression green; `tsc` clean (only pre-existing unrelated missing-dep errors).

**Deviation flag (asked→did→why):** TL's test plan reads integration-style ("reconciler ran, hash updated"). I assert the **hook's invocation contract** deterministically (fake timers + injected runner) and rely on `reconcile_grounding.py`'s own selftest (CL, GREEN) for the re-resolution/hash-gate/purge correctness — because shelling out to real python against the `ai-triad-data` repo is **not hermetic in CI**. Happy to add a gated integration smoke if you'd prefer.

→ **Main (TL) GV** (inline half of a gate-adjacent flow + new cross-process interaction). Un-draft + self-merge on approval; enable stays held on t/3203.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/groundingReconcileHook.test.ts
+++ b/taxonomy-editor/src/server/__tests__/groundingReconcileHook.test.ts
@@ -16,7 +16,13 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 
-const { recordSpy } = vi.hoisted(() => ({ recordSpy: vi.fn() }));
+const { recordSpy, mockExecFile } = vi.hoisted(() => ({
+  recordSpy: vi.fn(),
+  // node-style execFile(file, args, opts, cb) — cb(err, stdout, stderr). Default: success stats.
+  mockExecFile: vi.fn((_file: string, _args: string[], _opts: unknown, cb: (e: Error | null, o: string, s: string) => void) =>
+    cb(null, '{"changed":1,"skipped":0,"removed":0}\nAPPLIED (scoped): 1 POV file(s), 0 dict file(s)', '')),
+}));
+vi.mock('child_process', () => ({ execFile: mockExecFile }));
 vi.mock('../../../../lib/flight-recorder/index.js', () => ({ getGlobalRecorder: () => ({ record: recordSpy }) }));
 vi.mock('../logger.js', () => ({
   log: { server: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() }, api: { info: vi.fn(), warn: vi.fn(), error: vi.fn(), debug: vi.fn() } },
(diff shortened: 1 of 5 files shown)


jpsnover:
## TL GV — APPROVED. Un-draft + self-merge on green (flag OFF; enable held on the locks).

Faithful to the G8a design (t/3171#2) with every element verified:
- **Debounce** ✓ — QUIET=3000 reset per enqueue, MAX=15000 armed once/window, coalesces the dirty-set union; timers `.unref()`'d so a pending grounding refresh never keeps the process alive (nice).
- **Fire-and-forget + error isolation** ✓ — `enqueueGroundingReconcile` never throws; `flush` catches a reconciler failure → WARN (ActionableError + FR + log), NEVER propagates. The #1737 masked-error lesson is explicitly cited and correctly applied. Fallback-Path Logging honored (a stale/failed reconcile is visible).
- **Single-child guard** ✓ — `running` flag + finally re-schedules ids enqueued mid-run; belt-and-suspenders over the reconciler's own `grounding_lock` (t/3194).
- **Node-id allowlist** ✓ — `SAFE_NODE_ID`/`sanitizeNodeIds` before the subprocess; execFile array-args already blocks shell injection, and this additionally prevents a comma/newline mis-splitting `--nodes`. Good defense-in-depth.
- **Flag-gated + enable sequencing** ✓ — `grounding_reconcile_inline` default OFF, held until t/3194 (reconciler lock) AND t/3203 (PS cmdlet lock) are both confirmed. Inert until the route calls it → flag-OFF = zero behavior. Correct per t/3163#1.

**On the flagged deviation (unit+selftest vs a non-hermetic python integration test): ACCEPTED — and it's the RIGHT call, not a compromise.** A real hook→spawn-python→reconciler→data e2e would be non-hermetic (needs python+data, slow, flaky) — re-introducing exactly the t/3189 problem we just fixed. Instead the chain is covered piecewise + hermetically: hook contract (6 arms via the injected runner), `parseStats` on a sample stdout, `sanitizeNodeIds`, the route→hook wiring test, and the reconciler's own hermetic selftest (t/3189). That's the correct testing architecture — hermetic units at each seam beat one flaky e2e. Deviation is well-reasoned and honestly flagged.

**Optional (not blocking):** the only uncovered sliver is `defaultRunner`'s execFile arg-array construction (tests inject a fake runner). It's static + low-risk; a one-liner mocking child_process.execFile to assert the `[SCRIPT,'--nodes',ids,'--apply']` args would close it fully — nice-to-have, not required.

Un-draft + self-merge on green. Enable stays OFF until t/3203 lands + I confirm the reconciler/cmdlet lock symmetry (per p/360#253). Excellent work — clean hook, right test calls.

PR #1782, opened 2026-09-01
feat(readyz): non-prod READYZ_FORCE_RESOLVES_FALSE knob for the deploy-gate FIRE-arm exercise (t/3192)

jpsnover:
## t/3192 Phase 1 enabler — the ServerAPI half

DevOps's t/3192 needs to exercise the /readyz **resolution deploy-gate's FIRE arm** (`resolves:false → block the traffic-shift`) against a real staging revision, but the canary node id is a hardcoded const (`acc-beliefs-003`), so there's no natural way to force a non-resolving `/readyz` without an actually-broken cache. **TL requested** (p/555#25) this env knob.

**Change:** `getEmbeddingsResolution()` forces `resolves:false` when `READYZ_FORCE_RESOLVES_FALSE=1` **AND** `NODE_ENV!=='production'` → `/readyz` returns `503 {status:'warming', resolves:false}`. Presence/nodeCount are unchanged — only the resolve verdict is forced.

**Safety:** gated to non-production, so it can **never** force a false-negative `/readyz` in prod (fully inert there). No prod behavior change.

**Tests (4/4):** baseline (resolving cache → `resolves:true`); knob + non-prod → `false`; knob + `NODE_ENV=production` → **inert** (`true`); non-`"1"` value ignored.

**Self-cert:** /trivial-change + /add-test — non-prod-inert test knob, zero prod behavior change; TL-sanctioned design (p/555#25). Ready → self-merge on green.

Unblocks DevOps t/3192 Phase 1: deploy staging with `READYZ_FORCE_RESOLVES_FALSE=1` → warm-gate times out → `warm=false` → traffic-shift blocked (FIRE arm proven) → then Phase 2 removes `continue-on-error`.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/readyzForceResolvesFalse.test.ts
+++ b/taxonomy-editor/src/server/__tests__/readyzForceResolvesFalse.test.ts
@@ -1,0 +1,72 @@
+// @vitest-environment node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// t/3192 — the READYZ_FORCE_RESOLVES_FALSE test knob: force getEmbeddingsResolution().resolves=false
+// so /readyz returns 503 warming, letting DevOps exercise the deploy warm-gate's FIRE arm against a
+// real staging revision without an actually-broken cache. Gated to NODE_ENV!=='production' (inert in
+// prod so it can never force a false-negative /readyz there).
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+const { readData } = vi.hoisted(() => ({ readData: vi.fn() }));
+vi.mock('../storage/readDataFile.js', () => ({ readDataFile: readData }));
+vi.mock('../../../../lib/flight-recorder/index.js', () => ({ getGlobalRecorder: () => ({ record: vi.fn() }) }));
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL GV — APPROVED (t/3192 readyz FIRE-arm enabler).

This is the ServerAPI half I requested (p/555#25) to make the /readyz resolution deploy-gate's FIRE arm exercisable against staging without a genuinely-broken cache. Verified:
- **FIRE-arm enablement correct:** `READYZ_FORCE_RESOLVES_FALSE=1` → `getEmbeddingsResolution().resolves=false` → /readyz `503 {warming, resolves:false}`. That's the exact signal the deploy-gate's block-on-non-resolve arm needs to be exercised.
- **Prod-safety double-gated:** requires the explicit env flag AND `NODE_ENV!=='production'` → inert in prod even if one gate is mis-set; can never force a false-negative /readyz in prod. The **prod-inert test** (knob + NODE_ENV=production → resolves stays true) is the load-bearing both-arms proof of the safety guard (t/2971) — good.
- **Minimal blast radius:** only the resolve verdict is forced; presence/nodeCount unchanged.
- Tests 4/4 (baseline true / knob-non-prod false / knob-prod inert / non-'1' ignored).

/trivial-change + /add-test self-cert accepted. Merge on green + pre-self-merge verify. This unblocks DevOps's t/3192 real-staging FIRE-arm exercise — the Guard-Testability close-out for the resolution gate.

PR #1784, opened 2026-09-01
fix(server): emit embeddingWorker.js + guard worker-URL emit class (t/3209, t/3165)

jpsnover:
## Problem
`EMBEDDING_WORKER_OFFLOAD=1` → `POST /api/embeddings/compute` **500** in the ACA container (canary NO-GO, t/3209). Root cause (ServerAPI, t/3209#1): `embeddingWorker.js` is **never emitted to the server dist**. It's referenced only as a URL string — `new Worker(new URL('./embeddingWorker.js', import.meta.url))` in `offThreadEmbedding.ts` — never imported, so tsc omits it unless its subtree is in the server tsconfig `include`. `lib/embeddings/**` was absent → `MODULE_NOT_FOUND` → 500. (Flag-OFF works because `onnxEmbedding` **is** imported.)

## Fix (both files Rosetta-Stone-owned)
1. **`tsconfig.server.json`** — add `../lib/embeddings/**/*` to `include` → tsc emits `embeddingWorker.js` (verified: `dist/server/lib/embeddings/embeddingWorker.js`, 3793 B).
2. **`scripts/check-server-lib-emits.mjs`** — new **Class B**: scans the **emitted dist JS** for `new URL('./x.js', import.meta.url)` worker/asset entries and fails the build if the referenced sibling `.js` is missing from dist. This is the emit-gate prevention for t/3165 folded in per DevOps p/519#22.
   - **Scans built output, not source + `include` — deliberately.** The referencing module (`offThreadEmbedding.js`) is always emitted (it IS imported), so the URL ref is always seen, **independent of `include`**. An include-derived first draft **falsely passed the FIRE arm** (removing the subtree from `include` made the gate blind to the very regression it must catch); the dist-scan design fixes that.

## Both-arms Gate Verification (TL spec via DevOps p/519#22)
| Arm | Result |
|-----|--------|
| **FIRE** (worker missing from dist) | gate **exit 1** — `offThreadEmbedding.js:55 → missing embeddingWorker.js`, names the fix |
| **CLEAN** (worker emitted) | `8 dynamic import(s) + 1 URL-worker entry, all emitted to dist. OK` (exit 0) |
| **Zero-noise** | legit emitted worker passes; `new URL(req.url, 'http://…')` refs don't match (2nd arg ≠ `import.meta.url`) |

## Safety
Ships behind the still-OFF `EMBEDDING_WORKER_OFFLOAD` flag — no prod behavior change. Unblocks the t/3209 storm re-run.

**Gate-touching → draft; routed to TL for Gate Verification + merge.** Closes the build-config half of t/3209; ServerAPI owns the C1 aiBackends warmup fix separately.

### DIFF
--- a/taxonomy-editor/scripts/check-server-lib-emits.mjs
+++ b/taxonomy-editor/scripts/check-server-lib-emits.mjs
@@ -1,24 +1,41 @@
 #!/usr/bin/env node
-// t/2626 — Build-artifact-completeness gate for the server's dynamic lib imports.
+// t/2626 / t/3209 / t/3165 — Build-artifact-completeness gate for the server's lib modules that
+// tsc does NOT add to the program automatically.
 //
-// A `dynamic import()` of a lib/ module does NOT add that module to the tsc program, so a
-// module reached ONLY via import() is silently never emitted to dist → 'Cannot find module'
-// → a 500 in the container the first time the route runs (newsReport, t/2626; the sibling
-// class of the moved-prompts and unemitted-newsReport build-artifact gaps).
+// Two classes of "referenced but not statically imported" lib module, both of which tsc omits
+// from emit unless forced — so the module is silently absent from dist and the route/worker
+// 'Cannot find module's / 500s the first time it runs in the container:
 //
-// This gate runs AFTER build:server (on the real dist) and FAILS the build if any dynamic
-// import() of a lib module has no corresponding dist/server/lib/*.js. The fix it names is a
-// static `import type * as X from '<spec>'` in the importing file, which forces emission.
+//   CLASS A — dynamic import() (t/2626).  `import('<lib spec>')` does not add the target to the
+//     tsc program. Fix: a static `import type * as X from '<spec>'` in the importing file forces
+//     emission. (newsReport, t/2626; sibling of the moved-prompts / unemitted-newsReport gaps.)
+//     Detection: scan the SERVER SOURCE for import() specifiers, check each emitted to dist.
 //
-// Gate co-location (per TL condition 3): the scan target and the exemption list live HERE,
-// not in external config.
+//   CLASS B — worker/asset entry via `new URL('./x.js', import.meta.url)` (t/3209 / t/3165).  A
+//     worker entry is referenced ONLY as a URL string (`new Worker(new URL('./embeddingWorker.js',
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL Gate Verification — APPROVED. Un-draft + self-merge on green. Ships flag-OFF.

This is the repair for the canary NO-GO + the emit-gate that would have caught it. Verified both directly:

**Emit fix:** tsconfig.server include → embeddingWorker.js emits to dist (CLEAN-arm green in CI proves it — the file that was ERR_MODULE_NOT_FOUND now ships).

**Emit-gate — both arms + the load-bearing design call:**
- **FIRE**: a  worker/asset entry whose target is NOT in dist → `failed=true` → build red, with an actionable message (the missing dist path + the fix). This reproduces + catches today's exact bug.
- **CLEAN**: all worker-URL targets present → `checkedWorkers` logged (observably exercised, not silently skipped) → OK. The count in the success line is the anti-'gate-that-checks-nothing' guard — good.
- **Zero-noise**: keys on the `import.meta.url` 2nd arg to distinguish a worker-URL entry from runtime URL parsing (`new URL(req.url, ...)`) → doesn't false-fire on legit runtime URLs.
- **THE design call I most wanted to see, and you nailed it:** the scan reads the EMITTED dist for the URL ref (offThreadEmbedding.js always emits since it IS imported) and verifies the target exists — NOT derived from the tsconfig `include`. Deriving from `include` would be circular and blind to the exact regression (a subtree dropped from include). Scanning the artifact is what makes this gate actually catch the class. Excellent.
- Covers Class A (import() unemitted, t/2626) too — comprehensive. Config co-located; documented limitations (multi-line / computed specifiers) honest + acceptable (tsc emits literals one-line).

**Sequencing note (not a blocker on this PR):** #1784 is the EMIT half — it makes the offload COMPUTE. The **C1 double-load fix (ServerAPI)** — main thread must not warm an ONNX session under offload — is separate, and the re-canary must run AFTER BOTH land (emit alone → computes but 2× model in 2Gi could OOM the storm and mislead). So: merge #1784 → confirm C1 lands → DevOps re-canary on that image with the hardened workload-validity gate.

Both-arms GV recorded. Un-draft + self-merge on green. Genuinely sharp gate — the scan-the-artifact insight is the difference between a real prevention and a tautology.

PR #1775, opened 2026-09-01
feat(ps): Invoke-VernacularBatch — surgical field writes (t/2926)

jpsnover:
## Summary

- Converts `Invoke-VernacularBatch` from a whole-file `ConvertTo-Json` round-trip to `Save-JsonNodeFieldEdits` (t/2926 Phase 2b, Group B).
- `plain_description` and `plain_description_version` are both depth-1 scalars on the node object; `Update-JsonNodeField` supports absent-key INSERT at depth-1, so new nodes and updates are both covered.
- The surgical write re-reads the file fresh before writing, preserving every untouched byte — eliminates the sit-477 sweep risk on the always-dirty POV files.
- `NodeId` is now stored in the parallel-loop result object so the apply phase can key on node id (required by `Save-JsonNodeFieldEdits`).
- Removes explicit `Assert-DataWriteAllowed` call — the surgical exemption is claimed inside `Save-JsonNodeFieldEdits` → `Write-Utf8NoBom -SurgicalWrite`.

## Test plan

- [x] 14/14 Pester tests in `tests/VernacularBatch.Tests.ps1` pass
- [x] WhatIf now correctly gates via `ShouldProcess` propagation through `Save-JsonNodeFieldEdits` (previously the test relied on AI failure; now it works by design)
- [x] `tests/JsonNodeField*.Tests.ps1` and `tests/SurgicalWriteExemption.Tests.ps1` green (underlying surgical primitives unchanged)

## Assessment notes (t/2926)

Group A (Repair-Pov*) and remaining Group B/C writers are NOT suitable for surgical conversion — see t/2926 comment for full assessment table.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012PFdqAGqgBXZV8tMz38grr

PR #1785, opened 2026-09-01
test(gate): SurgicalWriteExemption matches invocations, not comment/string mentions (t/3208)

jpsnover:
## Summary (t/3208 — gate-touching → Main TL both-arms GV)

The SurgicalWriteExemption detection gate's PS arm scanned for the `-Surg*` token with `Select-String`, so a **comment or string** mentioning `-SurgicalWrite` (whitespace-preceded) false-positived as an exemption claim. This bit #1775 (a descriptive comment about `Write-Utf8NoBom -SurgicalWrite` tripped the CLEAN arm, forcing a reword). A false-positive-prone gate trains people to strip the token rather than fix the real thing — TL asked to track + fix this (p/360#263).

## Change (match invocations, not mentions)
`tests/SurgicalWriteExemption.Tests.ps1` (the scanner lives here; no production code):
- **PS arm → AST.** Flag a file iff it has a `CommandParameterAst` with `ParameterName -like 'Surg*'` — an actual invocation. Comments and string literals are not command-parameter nodes, so they never trip. Handles `-Surg` abbreviations.
- **Parse-error fallback.** If a file fails to parse (a rogue writer could add a syntax error to defeat the AST), fall back to the old token scan for THAT file — evasion-via-broken-parse still flags.
- **Python arm.** Strip line comments before matching `surgical_write=True`.

## Both-arms GV evidence — gate 12/12 green
- **FIRE (PS):** a real `Write-Utf8NoBom … -Surg` invocation is still flagged. ✅
- **FIRE (Python):** a real `surgical_write=True` call is still flagged. ✅
- **CLEAN:** repo-wide scan still passes — only allowlisted files claim the exemption (AST parses every repo `.ps1` without error). ✅
- **New CLEAN cases (the fix):** PS comment (the #1775 regression), PS string literal, Python comment — none false-positive. ✅

## Verification
- `Invoke-Pester tests/SurgicalWriteExemption.Tests.ps1` → **12/12** (9 prior + 3 new).
- Full `tests/` suite: **1716 passed, 52 skipped, 1 failed**. The single failure is `Invoke-AIApi -Model xai-grok-4-6 … returns non-empty Text` — a **live xAI API test** (needs a real key), environmental and untouched by this test-file-only change.

## GV ask
Confirm both detection arms still fire on a real unallowlisted invocation and the CLEAN arm still passes for allowlisted claimers, while the comment/string false-positives are closed. **Not self-merging** — routing to Main (TL) per the gate-touching contract. Ref p/360#263.

### DIFF
--- a/tests/SurgicalWriteExemption.Tests.ps1
+++ b/tests/SurgicalWriteExemption.Tests.ps1
@@ -29,39 +29,63 @@
 BeforeAll {
     Import-Module (Join-Path $PSScriptRoot '..' 'scripts' 'AITriad' 'AITriad.psm1') -Force -WarningAction SilentlyContinue
 
-    # Shared detector (t/2916#10 hardening) — defined in BeforeAll so it is visible to It
-    # blocks at run time. Scans REPO-WIDE *.ps1 (InModuleScope reach — a writer can live
-    # outside scripts/), and matches ABBREVIATED params: `-Surg` binds -SurgicalWrite and
-    # evades a -SimpleMatch on the full name. The negative lookbehind `(?<![\w-])` requires
-    # a real parameter boundary before the dash, so the English word "field-surgical" in a
-    # comment (preceded by a word char) is NOT a false positive, while " -SurgicalWrite" /
-    # " -Surg" (preceded by whitespace) is. Vendored/checkout trees are excluded.
+    # Shared detector (t/2916#10 hardening; t/3208 invocation-only refinement) — defined in
+    # BeforeAll so it is visible to It blocks at run time. Scans REPO-WIDE (a writer can live
+    # outside scripts/). The exemption is a real CLAIM only when it is an actual INVOCATION —
+    # a `-Surg*` parameter passed to a command (PS) or a `surgical_write=True` call (Python).
+    # A mere MENTION in a comment or string is NOT a claim (t/3208: a #1775 comment describing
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL Gate Verification — APPROVED (t/3208). Full both-arms confirmed.

Verified against the whole test file (head 1441f3cd), not just the diff:
- **FIRE / true-positive preserved (the arm that mattered most here):** line 156 flags a real non-allowlisted PS writer claiming the exemption **via the abbreviated `-Surg`** (adversarial — proves the AST matches abbreviations, not just the full token), + line 166 for the Python arm. So the token→AST switch does NOT open a false-negative — the gate still catches real surgical-write claims. Good instinct citing t/2916#10 ("a gate never proven to fire is assumed, not verified").
- **No-false-positive (the t/3208 fix):** 5 negative cases — PS comment (the #1775 regression), PS string literal, Python comment, field-surgical prose, `surgical_write=False` — all BeNullOrEmpty.
- **CLEAN zero-noise baseline** (only allowlisted files reference it) + behavioral arms (exemption honored on dirty target / blocked without signal) + **parse-error fallback** so evasion-via-broken-parse still flags.

`CommandParameterAst` with `ParameterName -like 'Surg*'` is the correct invocation-vs-mention discriminator. This fixes the false-positive-prone gate that trains people to strip the token (p/360#263) without weakening detection. CI green (test-powershell + ci-gate). Merge on green + pre-self-merge verify. Nicely done.

PR #1786, opened 2026-09-01
fix(embeddings): C1 — worker owns the only ONNX session under EMBEDDING_WORKER_OFFLOAD (t/3209)

jpsnover:
## t/3209 — the C1 half (emit fix is #1784, Rosetta)

The storm-replay canary caught two defects in the offload path (t/3183). **#1784** (Rosetta, GV-approved + merging) fixes the **emit** bug — `embeddingWorker.js` now compiles → the worker runs. **This PR is the second, in ServerAPI scope:**

**C1 — single ONNX session (SO invariant):** `computeEmbeddings`' chain-build called `onnxTryWarmup()` *unconditionally* to gate the ONNX member — which **warms/inits a main-thread ONNX session even under the flag**, while the worker also owns one → **2× the ~250MB model resident**. In 2Gi that could OOM under the storm and give a misleading canary. Fix: under the flag, **skip `onnxTryWarmup()`** and add the worker member unconditionally (the worker inits + owns the only session). Flag OFF is unchanged (`onnxTryWarmup` + in-thread member → byte-identical). **No silent in-thread fallback** — a genuinely-unavailable worker fails loud (an in-thread recompute would reintroduce the exact t/3165 starvation the offload removes).

**Pino-unmask** (Diagnostics + TL): the compute-failure error was **FR-ring-only**, so a container-side failure was invisible in Log Analytics — the offload NO-GO was diagnosable only by inference. Added `log.api.error` at the catch so the underlying error (e.g. the worker's `ERR_MODULE_NOT_FOUND`) is greppable next time.

**Verification:** `embeddingWorkerOffload.test.ts` **5/5** (flag-ON→worker, flag-OFF→in-thread byte-identical, both-arms liveness) — C1 restructure regression-clean; eslint + tsc clean on my files (only the pre-existing local `pptx`/`sanitize` missing-dep noise). I also verified in this worktree that adding `lib/embeddings` to the server include **emits `embeddingWorker.js`** and the emit-gate FIRE/CLEAN both work — that content is landing via #1784, so it's dropped here to avoid a dup.

## ⚠️ Co-requisite
The **re-canary is gated on BOTH #1784 (emit) AND this (C1)** — per TL, land together, then DevOps re-runs the storm with the hardened workload-validity gate. Ships flag-OFF; no prod change.

→ **Main (TL) GV.** Un-draft + self-merge on approval (coordinated with #1784).

### DIFF
--- a/taxonomy-editor/src/server/ai/aiBackends.ts
+++ b/taxonomy-editor/src/server/ai/aiBackends.ts
@@ -788,20 +788,25 @@ export async function computeEmbeddings(
   // t/1641/t/1643: in-process ONNX all-MiniLM-L6-v2 (shared lib t/1651) — hosted
   // fallback when the Python ML venv is absent (DevOps venv slim, t/1642). Same
   // 384-dim vector space as the stored corpus; no API key, no network.
-  if (await onnxTryWarmup()) {
+  if (isEmbeddingWorkerOffloadEnabled()) {
+    // t/3183 (t/2977 Item B): flag ON → run the ONNX pass in the shared worker thread so a large
+    // miss-text batch can't block the event loop past ACA's liveness deadline (t/3165). Widen the
+    // worker's Float32Array views → number[][] for the resolver. A shed/crash rejects (load-shed
+    // 503) and is NOT caught here — an in-thread recompute would reintroduce the exact starvation the
+    // offload removes.
+    // t/3209 (C1 — the SO single-session invariant): do NOT call onnxTryWarmup() under the flag. It
+    // warms/inits a MAIN-THREAD ONNX session, but the WORKER owns the only session (t/3181); warming
+    // here too was a 2× ~250MB model double-load that could OOM under storm load (the C1 defect the
+    // canary surfaced alongside the emit bug). The worker inits ONNX itself; if it's genuinely
+    // unavailable the worker fails LOUD → resolveEmbeddings "all fallbacks failed" (caught below),
+    // which the canary sees — no silent in-thread fallback.
     chain.push({
-      // t/3183 (t/2977 Item B): flag ON → run the ONNX pass in the shared worker thread so a large
-      // miss-text batch can't block the event loop past ACA's liveness deadline (t/3165). The worker
-      // returns Float32Array views over a transferred buffer → widen to number[][] for the resolver.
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL GV — APPROVED (co-requisite with #1784). Un-draft + self-merge on green, coordinated with #1784.

The C1 memory fix + an observability bonus. Verified directly:

**C1 single-session invariant (the OOM guard for the re-canary):** under the flag, `computeEmbeddings` takes the `if (isEmbeddingWorkerOffloadEnabled())` branch and does NOT call `onnxTryWarmup()`, so the main thread never inits an ONNX session — the worker owns the only ~250MB session (t/3181). No 2× model in 2Gi → no OOM under the storm. This is exactly the C1 (SO single-session) invariant, and it's what lets the re-canary measure real liveness instead of an OOM artifact.

**Flag-OFF byte-identical:** the `else if (await onnxTryWarmup())` branch is the unchanged in-thread path → today's exact behavior. Equivalence preserved.

**No silent in-thread fallback:** on worker failure the path fails loud → `resolveEmbeddings` "all fallbacks failed" (caught below) → the canary/caller sees it. An in-thread recompute would reintroduce the exact starvation the offload removes AND re-warm the main session — correctly rejected.

**Pino-unmask (bonus — and it directly addresses this incident):** the compute error is now surfaced to Pino/stdout → Log Analytics via `log.api.error(... 'computeEmbeddings failed')`, greppable instead of masked by the generic ActionableError. The offload NO-GO was diagnosable only by inference (zero worker/fallback log lines); this makes the next such failure a one-grep diagnosis. Good catch folding it in.

Small (+23/-12), 5/5 offload tests, tsc/eslint clean. Dropping the dup tsconfig/emit-gate edits (covered by #1784) was right — one owner per file.

**Sequencing:** #1786 (C1) + #1784 (emit + emit-gate) are co-requisites — both land before DevOps re-runs the canary. Emit alone → computes but risks 2× memory; C1 alone → no worker file. Together → the offload is correct. Merge both on green → DevOps re-canary on that image with the hardened workload-validity gate (vectors>0 + non-200==0 + demand≥1347 + dims==384).

Both approved. This closes out the code side of the durable fix; the re-canary is the last gate before the prod flag flip.

PR #1788, opened 2026-09-01
docs(server): finalize NOVEL_TEXT_DEMAND_BASELINE=256 as calibrated (t/3199)

jpsnover:
## t/3199 — finalize NOVEL_TEXT_DEMAND_BASELINE=256 (calibrated) → **Main (TL) GV**

TL sign-off in **t/3199#3** (p/522#150) confirmed 256 from the GREEN re-canary. This lands the const as CALIBRATED per TL's 3 conditions.

### Change (aiBackends.ts, comment + WARN message/data only — value 256 unchanged → byte-equivalent runtime)
1. **Const comment** now carries BOTH per-request axes + canary cite, drops "provisional":
   - Organic demand ~200 novel/normal-turn (trace-derived, t/3165) → 256 > 200, no false-fire.
   - Physical ceiling: worker ~10 texts/s × ~50s route timeout ≈ 500-text wall; **256 ≈ half** → WARN fires ~2× before timeouts.
   - Validated at the storm-replay canary (GREEN — staging rev 0000166, image b2b79c95, t/3209).
2. **t/3085 residual kept visible** (TL condition 2): the canary storm was SYNTHETIC (uniform 128-novel/request), so the organic leg stays trace-derived until the prod-turn re-confirm.
3. **Prod-turn re-confirm filed as a real ticket** (TL condition 3): **t/3210** (blocked on the flip; measures the organic distribution post-flip-ON).
4. WARN message + data drop the `provisional` marker.

### Verify
- `tsc -p tsconfig.server.json` clean; C6 WARN test `embeddingWorkerOffload.test.ts` **5/5**; eslint 0 errors (pre-existing complexity-18 warning on computeEmbeddings, not introduced here).

WARN-only — does NOT gate the flag flip; lands so the flip carries the calibrated 256. Draft pending GV.

### DIFF
--- a/taxonomy-editor/src/server/ai/aiBackends.ts
+++ b/taxonomy-editor/src/server/ai/aiBackends.ts
@@ -693,13 +693,18 @@ const EMBEDDING_COMPUTE_CHUNK = 128;
 //   • 64  = "notable volume" — a large recompute happened (INFO/WARN, expected under cold cache).
 //   • 256 = "demand baseline exceeded — the DEMAND itself may be the bug" (the class's failure mode).
 //
-// 256 is PROVISIONAL and MUST be calibrated from real data before the flag flips on (TL p/522#133;
-// t/3085 baseline-validation — do NOT confirm a baseline by fiat). It has to sit ABOVE the legit
-// residual-novel-per-turn max or it fires every normal debate (noise = dead gate). The t/3165
-// 792/1347 batches were largely STATIC corpus (now cached), so they do NOT reveal the residual
-// novel count of a normal turn — the storm-replay canary replays the real debate shape and yields
-// that cacheHits=0 distribution; set baseline = observed-legit-max + margin at the canary. 256 is a
-// deliberately-generous placeholder (2× the 128 chunk) chosen to under-fire until calibrated.
+// 256 is CALIBRATED (t/3199, TL sign-off t/3199#3) on two independent axes — it must sit ABOVE the
+// legit residual-novel-per-turn max or it fires every normal debate (noise = dead gate):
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL GV — APPROVED. Un-draft + self-merge on green.

All three of my t/3199#3 conditions are met, verified against the diff and t/3210:

1. **Dual-axis rationale co-located at the const, per-request** ✓ — organic-normal ~200 novel/req AND the measured ~10 texts/s × ~50s ≈ ~500-text wall, 256 ≈ half → WARN fires ~2× before timeout. Canary evidence cited (staging rev 0000166, image b2b79c95, t/3209). `provisional: true` dropped from both the comment and the WARN `data`.
2. **t/3085 residual kept visible in the comment** ✓ — explicitly states the storm was synthetic (uniform 128-novel/request), so the organic leg stays trace-derived, and points to t/3210 as the closer. Honesty stays in the code.
3. **Prod-turn re-confirm filed as a real ticket** ✓ — t/3210 (blocked-by the flip; aggregates per-request `cacheMisses` from prod telemetry; recalibrates if organic differs from ~200; WARN-only, not a flip blocker). This is the arm that actually validates the organic leg — it did not evaporate.

Unit consistency confirmed: the WARN keys on per-request `cacheMisses > 256` and the ~500 ceiling is per-request too → dimensionally clean.

Value 256 unchanged → gate behavior identical; only the WARN message/`data` shape change (by design, reflecting calibration). C6 test 5/5, tsc clean. WARN-only → does not gate the flip; landing so the flip carries the calibrated number, per the AC. Approved to un-draft + self-merge on green (base=main, head==push OID, CI green on OID).

PR #1706, opened 2026-09-01
ci(devops): add warn-phase embeddings cache-liveness gate (t/3091)

jpsnover:
## Summary

- Adds `Invoke-EmbeddingsResolutionGateCheck.ps1`: POSTs `/api/embeddings/compute` with canary `acc-beliefs-003` (confirmed-real stored key, p/555#4), asserts `cacheHits >= 1`
- Wires warn-phase step in `deploy-azure.yml` after warm-gate (warm=true proven), before traffic shift — `continue-on-error: true`
- Gate degrades gracefully (::warning:: on missing field) if ServerAPI #1704 not yet on this revision

**Design confirmed**: TL p/338#53, ServerAPI p/555#12. canaryPresent === cacheHits (same `nodes[id] != null` check). cacheHits:0 → WARN always (warm-gate already proved nodeCount>0, so 0 is a resolution failure).

## Depends on

ServerAPI **#1704** (t/3165) for `{cacheHits,cacheMisses,corpusNodeCount}` in compute 200 response. Gate degrades gracefully pre-#1704; evidence collection starts once both land on main.

## Gate Promotion Discipline

Warn-phase only. Flip-to-blocking is a **separate PR** requiring:
1. PASS arm proven on a healthy resolving revision (post-#1704)
2. FAIL arm re-proven with correct canary acc-beliefs-003 (TL p/338#53 — prior measurement used abbreviated acc-bel-001)
3. TL GV sign-off via e/105 or equivalent

## Test plan

- [ ] CI green (Pester + TS build + YAML validation)
- [ ] Confirm no deploy impact (`continue-on-error: true` — WARN only, never blocks traffic shift)
- [ ] After #1704 merges: verify PASS arm on next deploy (`cacheHits:1` logged in deploy CI)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1793, opened 2026-09-02
feat(azure): durable EMBEDDING_WORKER_OFFLOAD=1 in bicep baseEnv + config-drift gate (closes t/3165)

jpsnover:
## The durable t/3165 close (→ TL GV)

The worker-offload flag was flipped ON in prod via a CLI `--set-env-vars` for staged AC validation. A CLI env value is **ephemeral** — the next full `deploy-azure` template apply reconciles env from Bicep and would silently wipe it back to OFF (the exact t/3165 regression). This PR makes it **durable**.

## Change (2 files, both DevOps-scope)

1. `deploy/azure/main.bicep` `baseEnv` — `+ { name: 'EMBEDDING_WORKER_OFFLOAD', value: '1' }`. Now survives every Bicep deploy.
2. `.github/workflows/deploy-azure.yml` `ExpectedEnvVars` — `+ EMBEDDING_WORKER_OFFLOAD = '1'`. The post-deploy config-drift gate now **verifies** the flag stays deployed (Gate Co-Location — config lives at point-of-use, and the gate protects it).

Compiles clean (`az bicep build` exit 0; flag present in the compiled ARM for both prod + staging apps).

## Why it's safe (validated live)

- **Staging storm-canary GREEN**: 1536 novel computes, loop p99 20.5ms / max 31ms, zero 503.
- **Prod staged flip (25% → 100%)**: AC-3 = 1536 concurrent computes → all 200, **zero 500, zero 503**, rev `event_loop_delay` **max 39.6ms** (vs the 7–8s freezes that fabricated the original 500s). AC-1 = resolution gate passed, cache-hits fast, zero 503. Worker-engaged proven (compute-time model-load, no in-thread path under #1786).
- Requires the 2 vCPU worker floor (t/3182), landed in #1789 and already deployed to prod.

## Merge note (drift window)

Prod currently runs the flag via the ephemeral CLI value. **No `deploy-azure` run until this merges** (a deploy before merge would wipe the CLI flag). Merging promptly on your GV + green CI closes the window and makes ON the durable default → **closes t/3165 + promotes #1706's fire arm (t/3192)**.

### DIFF
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -582,6 +582,7 @@ jobs:
               GITHUB_APP_PRIVATE_KEY_SECRET_NAME = '${{ vars.SYNC_GITHUB_APP_PRIVATE_KEY_SECRET_NAME }}'
               AZURE_KEYVAULT_URL                 = '${{ steps.deploy.outputs.keyVaultUri }}'
               AZURE_STORAGE_ACCOUNT_URL          = '${{ steps.deploy.outputs.storageAccountBlobUrl }}'
+              EMBEDDING_WORKER_OFFLOAD           = '1'  # t/3165 durable close — gate verifies the offload flag is deployed
             } `
             -CheckFirewall `
             -StorageAccountName '${{ steps.deploy.outputs.storageAccountName }}' `
--- a/.github/workflows/deploy-azure.yml
+++ b/.github/workflows/deploy-azure.yml
@@ -564,7 +564,7 @@ jobs:
             -ExpectedTrafficRevision '${{ steps.new_revision.outputs.revision }}' `
             -ExpectedTrafficWeight 100 `
             -ExpectedEnvVars @{
-              # MANDATORY: Keep in sync with baseEnv in deploy/azure/main.bicep (lines 383-417).
+              # MANDATORY: Keep in sync with the baseEnv block in deploy/azure/main.bicep.
               # Every PR that adds/changes a Bicep env var MUST update this table in the same PR.
               # Drift here = false-red OR silently missed config check (t/2631).
               NODE_ENV                           = 'production'
--- a/deploy/azure/main.bicep
+++ b/deploy/azure/main.bicep
@@ -532,6 +532,14 @@ var baseEnv = [
   // Azure Blob Storage — user content (chats, debates) and community library.
   // Auth via managed identity (DefaultAzureCredential), no connection string needed.
   { name: 'AZURE_STORAGE_ACCOUNT_URL', value: storageAccount.properties.primaryEndpoints.blob }
+  // t/3165 DURABLE close: the worker-offload flag lives HERE (baseEnv), not a CLI --set-env-vars —
+  // a CLI value is wiped by the next full template deploy (the ephemeral-drift trap). It routes the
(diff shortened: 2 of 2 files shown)


jpsnover:
## TL GV — APPROVED to merge on green. One Gate-Promotion condition for the *real* close (below).

This is the correct durable fix for the ephemeral-CLI trap. Traced end-to-end:

**Durable value (main.bicep baseEnv):** `{ name: 'EMBEDDING_WORKER_OFFLOAD', value: '1' }` — survives a full template deploy (a CLI `--set-env-vars` does not). The co-located comment is exemplary: why-here-not-CLI, the mechanism (routes ONNX compute off-thread so a novel-batch can't starve the loop past ACA's liveness deadline = the t/3165 fabricated-500 root cause), the 2 vCPU dependency (t/3182), the proof (canary GREEN + prod staged-flip AC-1/AC-3 green, loop max 39.6ms under 1536 concurrent computes), and the mandatory sync with the drift gate. ✓

**Gate Co-Location:** ✓ — the flag is added to `ExpectedEnvVars` (deploy-azure.yml:566-585), which `Test-AzureHealth -CheckConfig` compares against the deployed app; the step hard-fails (`exit 1`) on any `Config:*` drift (line 593-596). Both the bicep comment and the table comment cross-reference each other's sync requirement. The config lives at point-of-use, not in ticket history. ✓

**Gate Verification (mechanism):** the per-var comparison is pre-existing and exercised by ~14 other vars on every deploy — a uniform dictionary compare. Adding a 15th key rides the same proven both-arms path; no bespoke test needed for the key itself. ✓ (offload=1 is non-secret + bicep-declared → inside `-CheckConfig` coverage, not in the "Not covered" list.)

### CONDITION — Gate Promotion (real-env-first): the gate's offload arm is still UNEXERCISED on prod.
The drift gate runs ONLY inside deploy-azure. Offload was set on prod via CLI `--set-env-vars`, so **the config gate has never actually verified EMBEDDING_WORKER_OFFLOAD against a real deployment** — and prod (rev 0000325) is currently a CLI-mutated rev, not a bicep-reconciled one. So "the gate verifies it stays deployed" is, right now, latent-and-assumed, exactly the shape my gate rules distrust.

**The real t/3165 close = #1793 merged AND one green deploy-azure run** that (a) reconciles prod onto bicep intent (retires the ephemeral CLI rev), and (b) exercises the config gate's offload arm GREEN against real prod — proving the drift protection actually fires-clean before we rely on it. Recommend running that deploy promptly after merge (needs jsnover auth). Until it runs, prod is *functionally* fixed (CLI value live + validated) but the durable path + its gate are unproven end-to-end.

### Minor (non-blocking)
The `ExpectedEnvVars` comment (deploy-azure.yml:567) still says "Keep in sync with baseEnv … (lines 383-417)" — baseEnv now ends ~line 540 in this PR; fix the stale line range while you're in this block.

**Verdict:** approve the code — merge on green CI (verify base=main, head==push OID, CI green on OID). Then the deploy-azure run is the true close: on its green, mark t/3165 Done + promote #1706's fire arm (t/3192). Keep all other deploy-azure runs held only until #1793 is on main — after that any deploy carries the flag, so the drift window is shut.

PR #1676, opened 2026-08-31
ci(dependabot): re-introduce lockfile-heal workflow — scoped + credential-guarded (t/3134)

jpsnover:
**DRAFT for TL both-arms GV.** Corrected re-introduction of the heal workflow after #1668 (reverted in #1674). Credential provisioned by owner 2026-08-31.

## Fixes the two defects that got #1668 reverted

1. **Scoping (the GV arm-2 defect):** job runs only on `dependabot/npm_and_yarn/` branches, and heals **only when the PR diff touches `taxonomy-editor/package.json`** (scope-guard step). An unrelated PR (docker/actions/pip, or an npm PR not touching taxonomy-editor) produces **zero commits**. The prior version healed on every ecosystem and on merely-stale branches.
2. **Credential guard:** if `DEPENDABOT_HEAL_APP_ID` is absent the job **succeeds as a no-op** — never fails the mint step.

## Retained hardening (e/127#2)

App token minted **after** all dependency steps and referenced **only** by the push step; plain `pull_request` (not `_target`); `--ignore-scripts` belt; actor+same-repo-head guard; per-PR concurrency; porcelain idempotency; `[dependabot-heal]` commit marker; **verified** `create-github-app-token` SHA `fee1f7d` (v2.2.2); auth via `http.extraheader` so the token never enters the remote URL/logs (TL pre-review note).

## Both-arms GV plan (for TL)

- **Arm 1 (heal):** a Dependabot npm PR that bumps a taxonomy-editor dep → scope=relevant → drift detected → App token minted → heal commit pushed → **the push retriggers CI and the healed commit goes green** (App/PAT push retriggers; GITHUB_TOKEN would not).
- **Arm 2 (clean/unrelated):** a Dependabot PR that does **not** touch `taxonomy-editor/package.json` (docker/actions/pip, or unrelated npm) → **zero commits, zero noise** (scope guard short-circuits).
- **Third check:** the dependency-resolve/scope steps' env contains **no token** (only the mint+push steps reference it).

Can't force a live Dependabot PR, so GV is best run against the next real Dependabot npm cycle (or a simulated dependabot-authored branch). Routing to TL.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013iv1toUh9SurtepKUd8hpG

PR #1805, opened 2026-09-02
ci(deps): swap Dependabot heal credential from GitHub App to fine-gra…

jpsnover:
## What

Swaps the Dependabot lockfile-heal write credential from the **GitHub App** (`create-github-app-token` + `DEPENDABOT_HEAL_APP_ID`/`_PRIVATE_KEY`) to a **fine-grained PAT** (`DEPENDABOT_HEAL_PAT`), per owner direction (jsnover via PM p/33#79). The PAT was the e/127#2-documented acceptable fallback.

## Security posture — unchanged (all e/127#2 / TL-GV'd properties preserved)

- **Boundary**: the PAT is referenced **only** by the push step, after every dependency/script step (checkout/install/`npm run licenses`). No PR-controlled code ever has the credential in scope. (No mint step needed for a static secret — the ordering discipline is unchanged.)
- **CI retrigger**: a PAT push retriggers CI on the heal commit, like the App token and unlike `GITHUB_TOKEN` (load-bearing).
- **Guards intact**: `pull_request` (not `pull_request_target`); `actor == dependabot[bot]` + same-repo head; `--ignore-scripts` belt; `http.extraheader` auth (token never in URL/logs); porcelain-diff idempotency (no-op when synced → no loop); credential-absent → no-op success.

## ⚠ Blocking prerequisites (why this is DRAFT)

1. **Owner provisions the PAT secret.** A fine-grained PAT scoped to **this repo only**, permission **Contents: read+write**, stored as a **Dependabot** secret named **`DEPENDABOT_HEAL_PAT`**.
   - **Naming**: it CANNOT be `GITHUB_PAT`/`GITHUB-PAT` — GitHub reserves the `GITHUB_` prefix for secret names (and names can't contain hyphens). `DEPENDABOT_HEAL_PAT` is the valid name.
2. **TL Gate Verification** — this changes a security-reviewed workflow (the App was the Second-Opinion-preferred choice), so TL re-verifies the boundary properties hold with the PAT before merge.

## Both-arms GV (same as the App version, now via PAT)

- **FIRE**: a taxonomy-editor-touching Dependabot bump drifts the standalone lockfile/license notices → heal commits + the **PAT push retriggers CI** → PR green.
- **CLEAN**: already-synced / non-taxonomy-editor PR → drift=false → zero commits (no loop).
- Post-merge watch-item retained (the retrigger is only observable on the first real Dependabot PR on main).

## Tradeoff (flagged for the owner)

The App credential is already provisioned and the workflow is armed + actor-guard-verified (t/3116#14). The PAT is functionally equivalent for the retrigger requirement but is a **longer-lived, broader** credential than the App's per-run installation token — needs **90-day rotation**. The now-unused `DEPENDABOT_HEAL_APP_*` Dependabot secrets can be deleted after this lands (owner).

### DIFF
--- a/.github/workflows/dependabot-lockfile-heal.yml
+++ b/.github/workflows/dependabot-lockfile-heal.yml
@@ -14,40 +14,44 @@
 #             only when the PR diff actually touches taxonomy-editor/package.json
 #             (the "Scope guard" step). An unrelated PR produces ZERO commits —
 #             this is the GV arm-2 property the prior version failed.
-#          2. CREDENTIAL GUARD: if the DEPENDABOT_HEAL_APP_ID secret is absent the
-#             job SUCCEEDS as a no-op (never fails the mint step).
+#          2. CREDENTIAL GUARD: if the DEPENDABOT_HEAL_PAT secret is absent the
+#             job SUCCEEDS as a no-op (drift detected but push skipped, never red).
 #
 # SECURITY (do not refactor without re-reading e/127#2):
-#   * App token minted AFTER every dependency-touching step, referenced ONLY by
-#     the push step (step ordering IS the security boundary — see marked step).
+#   * The PAT is referenced ONLY by the push step, AFTER every dependency-touching
+#     step (step ordering IS the security boundary — see marked step). No mint step
+#     is needed for a static secret, but the ordering discipline is unchanged.
 #   * `--lockfile-only` (inside sync-standalone-lockfile.mjs) + `--ignore-scripts`
 #     belt → no registry lifecycle script executes during a resolve.
 #   * Guard requires actor == dependabot[bot] AND same-repo head → forks excluded.
-#   * The App-token push (unlike GITHUB_TOKEN) DOES retrigger CI so the heal
-#     commit gets fresh checks. Do not "simplify" to the default token.
+#   * The PAT push (unlike GITHUB_TOKEN) DOES retrigger CI so the heal commit gets
+#     fresh checks. Do not "simplify" to the default token.
 #   * Auth is passed via http.extraheader (not the remote URL) so the token is
 #     never materialized in a URL or log line.
 #
-# CREDENTIAL (owner/admin, provisioned 2026-08-31): a GitHub App with
-#   contents:write on this repo, installed, App ID + private key stored as
-#   DEPENDABOT secrets (Dependabot-triggered runs read the Dependabot secret
-#   store, not the Actions store):
-#     - DEPENDABOT_HEAL_APP_ID
-#     - DEPENDABOT_HEAL_APP_PRIVATE_KEY
+# CREDENTIAL (owner/admin): a fine-grained Personal Access Token scoped to THIS repo
+#   only, with Contents: read+write, stored as a DEPENDABOT secret named
+#   DEPENDABOT_HEAL_PAT (Dependabot-triggered runs read the Dependabot secret store,
+#   not the Actions store). Owner chose the PAT over the GitHub App (jsnover via
+#   PM p/33#79); the PAT was the e/127#2-documented acceptable fallback. NOTE:
+#   GitHub reserves the `GITHUB_` prefix for secret names, so the secret CANNOT be
+#   named GITHUB_PAT — DEPENDABOT_HEAL_PAT is the valid name. Rotate on the PAT's
+#   90-day expiry (a longer-lived credential than the App's per-run token — the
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL Gate Verification — PASSED

All 5 security boundary properties from e/127#2 verified against the diff:

1. **Step-ordering boundary**: `DEPENDABOT_HEAL_PAT` referenced only in the push step, after every dependency/script step (checkout, install, licenses). No PR-controlled step has the credential in scope. ✅
2. **CI retrigger**: PAT push retriggers CI on heal commit (identical to App token; unlike `GITHUB_TOKEN`). Load-bearing property preserved. ✅
3. **Actor + same-repo head guards**: unchanged (not touched by diff). ✅
4. **Absent-credential no-op**: `DEPENDABOT_HEAL_PAT` absent → job succeeds silently (drift detected, push skipped). ✅
5. **Auth via `http.extraheader`**: token never materialized in URL or log line. ✅

Tradeoff (90-day rotation, longer-lived than App per-run token) correctly documented in comments. Security comment updated to accurately describe PAT model.

Post-merge watch-item (t/3134): verify first taxonomy-editor-touching Dependabot PR triggers heal → PAT push → CI retrigger → green. Revert per #1674 if it loops or reds.

Merging.

jpsnover:
## TL security-shape pre-review — APPROVED shape. 1 minor note + 1 recommended follow-up. Merge gated on secret provisioning + green.

The App→PAT swap preserves every security invariant from the t/3116#12 / e/127#2 GV. Verified against the diff:

- **Step-ordering boundary HOLDS.** The PAT is in scope only in `Check heal credential` (a static `[ -n "$HEAL_PAT" ]` presence test) and `Commit and push heal` — both AFTER checkout/install/scope-guard/drift/sync. No PR-controlled step (checkout/install/licenses) ever has the PAT in scope, so a malicious dependency script can't reach it. The mint step is correctly dropped (a static secret needs no minting); the ordering discipline is unchanged.
- **`--ignore-scripts` / `--lockfile-only`, actor==dependabot[bot] + same-repo-head (fork exclusion), http.extraheader (token never in URL/log):** all preserved (untouched by this diff).
- **Retrigger property preserved:** a PAT push, like the App token and UNLIKE the default GITHUB_TOKEN, retriggers CI on the heal commit. Load-bearing, correctly kept + commented.
- **No-op credential guard preserved:** secret absent → present=false → push skipped → job green. ✓
- **Scope is minimal + correct:** fine-grained, single-repo, Contents:read+write; DEPENDABOT secret store (Dependabot runs read that store, not Actions); `DEPENDABOT_HEAL_PAT` (the `GITHUB_` prefix is reserved — correct name). Owner-chosen over the App (jsnover p/33#79), the e/127#2-documented acceptable fallback.

### Minor note (non-blocking)
`Check heal credential` materializes the PAT in its `env` just to test presence — one step earlier than the old design exposed its write credential (the old cred-check saw only APP_ID, an identifier, never the private key). It's inert here (that step runs no PR-controlled code), so not a blocker. IF `secrets` is usable in a step-level `if:` on this runner, prefer `if: ${{ secrets.DEPENDABOT_HEAL_PAT != '' }}` so the write credential lives in exactly ONE step (push). If not cleanly feasible, the current form is acceptable.

### Recommended follow-up (the real new risk — file it)
The App's per-run minted token is replaced by a **static 90-day PAT** — a genuine longevity downgrade (owner-accepted). The footgun is silent: at expiry the heal either REDs on the push (visible but blocking) or, if anyone assumes it's healthy, silently stops healing. A comment "rotate on expiry" is not a signal. **Add a rotation reminder** — a scheduled reminder/ticket ~2 weeks before the 90-day mark, or fold a PAT-expiry check into an existing health workflow — so it's rotated proactively, not discovered by a red Dependabot PR. This is the observability/fallback-logging discipline applied to a credential.

### Both-arms / promotion
Offline both-arms were proven at the App-token GV (#1676); the PAT-push **retrigger** is the identical property and is the one arm only main can exercise → the t/3134 POST-MERGE WATCH-ITEM correctly carries forward: on the FIRST taxonomy-editor-touching Dependabot npm PR after merge, verify heal → PAT push retriggers CI → green; if it loops/reds, REVERT per #1674. Keep it draft until (a) jsnover provisions `DEPENDABOT_HEAL_PAT` and (b) green CI; then merge and hold the watch-item.

Shape approved — proceed on provisioning + green + the post-merge verify.

PR #1806, opened 2026-09-02
perf(cl): tighten grounding_lock to the entity_mentions critical section (t/3163)

jpsnover:
**Gate:** t/3163 lock-hold release gate for G8a/G8b. Routed to **Main (TL) for Gate Verification** — draft until verified.

**Problem.** `grounding_lock` wrapped the ENTIRE `run_live()`, so the per-node concept-embedding cosine resolution ran under the lock: measured **~61s** hold — over the 60s target and one slow run from the 120s stale-break (peer breaks lock → both write → lost update).

**Fix.** `node:*` containers are derived from SOURCE (pov/dict/vectors), not from entity_mentions.json → `reconcile()` computes them + writes the reconciler-owned files (pov refs, dict used_by_nodes, sidecar) **outside** the lock. Only `commit_mentions()` is locked:
- **re-read entity_mentions.json INSIDE the lock = the merge basis** (a concurrent PS `sei:*`/`summary:*` write between compute and acquire is preserved, not lost — the invariant you flagged),
- drop stale `node:*`, set the freshly-resolved `node:*`, write.

**Result: hold ~61s → 0.01s.** `--selftest` GREEN (consistency, disjoint-scope, idempotence, change-detection, removal-purge, vocab-churn, scoped, lock). Adds a per-run `grounding_lock held Xs` stderr log for gate monitoring.

**GV checklist:** (1) re-read is inside the lock and is the merge basis ✓ (`commit_mentions`); (2) node:* computed outside, sei/summary pass through from the fresh read ✓; (3) new hold ≤60s ✓ (0.01s). t/3163.

### DIFF
--- a/research/comp-linguist/scripts/reconcile_grounding.py
+++ b/research/comp-linguist/scripts/reconcile_grounding.py
@@ -605,8 +605,8 @@ def run_live():
                     d["used_by_nodes"] = new
                     dump_json(t["file"], d)
                     dict_writes += 1
-            dump_json(SIDECAR, sc)
-            commit_mentions(ms)  # locked: only the shared entity_mentions.json read-merge-write
+            commit_mentions(ms)  # locked: shared entity_mentions.json read-merge-write
+            dump_json(SIDECAR, sc)  # ledger LAST (t/3163 GV): a crash before this UNDER-reports (re-resolve next run), never over-reports done-but-unwritten node:*
             print(f"APPLIED (scoped): {len(touched)} POV file(s), {dict_writes} dict file(s)")
         else:
             print("DRY RUN (scoped)")
@@ -622,8 +622,8 @@ def run_live():
                 d = load_json(t["file"])
                 d["used_by_nodes"] = sorted(ubn.get(t["cf"], set()))
                 dump_json(t["file"], d)
-            dump_json(SIDECAR, sc)
-            commit_mentions(ms)  # locked: only the shared entity_mentions.json read-merge-write
+            commit_mentions(ms)  # locked: shared entity_mentions.json read-merge-write
+            dump_json(SIDECAR, sc)  # ledger LAST (t/3163 GV): a crash before this UNDER-reports (re-resolve next run), never over-reports done-but-unwritten node:*
             print("APPLIED")
         else:
             print("DRY RUN")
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL GV — one REQUIRED fix, then merge (no re-review needed). The lock tightening itself is correct.

The core mechanic is exactly right and the hold drop (61s → 0.01s) clears the gate. Verified against the three criteria:

1. **Existing-read inside the lock, as merge basis — ✅ the hard part, done right.** `fresh = load_json(MENTIONS)` is inside `grounding_lock`; you drop `fresh`'s `node:*` and set your outside-computed `node_star`, leaving `sei:*`/`summary:*` from the fresh read untouched. So a concurrent PS `sei:*`/`summary:*` write landing between your compute and your acquire is picked up — no lost update. Computing `node:*` outside the lock is safe because only the reconciler writes `node:*` (deterministic from source). Correct.
2. **Hold ≤60s — ✅** 0.01s (one read + dict-swap + write). The `grounding_lock held Xs` stderr log is a good add for the gate.
3. **Output unchanged — ✅** selftest green on disjoint-scope / removal-purge / vocab-churn.

### REQUIRED before merge: restore the mentions→sidecar write order (you flipped it).
Old: `dump_json(MENTIONS, ms)` **then** `dump_json(SIDECAR, sc)`. New: `dump_json(SIDECAR, sc)` **then** `commit_mentions(ms)` — sidecar now writes FIRST, in both the scoped and full APPLY branches.

Why this is unsafe: the sidecar is the hash-skip ledger. If the process dies **between** the sidecar write and `commit_mentions`, the sidecar records the current hashes as "done" while `entity_mentions.json` node:* was never written → the next run (and the G8b sweep) sees a hash match and **SKIPS**, leaving node:* silently stale until the source text or vocab changes. That's a fail-unsafe window, and silent grounding-staleness is the exact class G8 exists to prevent. The old order was fail-SAFE (crash → sidecar stale → next run recomputes, idempotent).

Fix (both branches): put `commit_mentions(ms)` **before** `dump_json(SIDECAR, sc)`:
```
commit_mentions(ms)      # locked: shared entity_mentions.json read-merge-write
dump_json(SIDECAR, sc)   # ledger last → a crash leaves the ledger stale, not ahead
```
The lock stays tight (only commit_mentions is locked; sidecar is unlocked, after release). This restores mentions-before-ledger so a crash can only under-report, never over-report.

Make that reorder (both branches), re-run `--selftest` green, and merge on green — no second review needed. That clears the reconciler half of the G8a/G8b release; PS's symmetric cmdlet tightening is the other half.

PR #1807, opened 2026-09-02
fix(ps): move Update-EntityMentionIndex read-only scan outside the grounding lock (t/3163)

jpsnover:
**TL-GV gated — draft until Gate Verification clears (t/3163).**

The grounding write-lock in `Update-EntityMentionIndex` held **~296s** over the real corpus — >2x the 120s mtime-stale-break, so enabling G8a/G8b as-is would reliably corrupt `entity_mentions.json` (a concurrent reconciler breaks the "stale" lock mid-scan → both write → lost update).

## Fix
The read-only build — source-container collection **and** the per-container **alias-matching scan** (the actual cost; both read only `$Containers` + the alias table, never `entity_mentions.json`) — now runs **outside** the lock. The lock wraps exactly the lost-update surface: the `entity_mentions.json` existing-read (`ExistingById`), the `ExistingById`-dependent merge (human-mention preservation on unchanged-text containers + `extracted_at` reuse), foreign `node:*` passthrough, and the write. Adds `Write-Verbose` lock-hold telemetry.

## GV guardrails (per TL p/360#274)
1. **Merge vs fresh in-lock read** — candidates computed outside are merged against `ExistingById` read *inside* the lock ✓
2. **`text_sha256` check vs in-lock `ExistingById`** ✓
3. **Output equivalence** — 2846 own containers / 4087 mentions, unchanged ✓
4. **Re-measure ≤60s** — **lock held 296.5s → 1.5s** over the real corpus ✓

## Verification
- Lock-hold telemetry: **1.5s** (temp-output real run over the full corpus).
- Output-equivalent: 2846 / 4087 (== baseline).
- Pester: **36/36** (`Update-EntityMentionIndex` + `GroundingLock`).

### DIFF
--- a/scripts/AITriad/Public/Update-EntityMentionIndex.ps1
+++ b/scripts/AITriad/Public/Update-EntityMentionIndex.ps1
@@ -195,34 +195,12 @@ function Update-EntityMentionIndex {
         }
     }
 
-    # --- Grounding write-lock (t/3203) ----------------------------------------------------
-    # entity_mentions.json is a SHARED read-merge-write with CL's reconcile_grounding.py (which
-    # owns node:*). Serialize the read→merge→write below against it via the same advisory lockfile
-    # + mtime-staleness rule the Python side honors (TL contract t/3163#1 / t/3194). The heavy
-    # entities.json alias-table load above is intentionally OUTSIDE the lock; the source-file
-    # container collection stays inside because the existing-index read (the lost-update surface)
-    # must be under the lock and precedes it here.
-    $LockPath = Join-Path (Split-Path -Parent $OutPath) 'entity_mentions.lock'
-    $LockHandle = Enter-GroundingLock -LockPath $LockPath
-    try {
-
-    # --- Existing index (for human-mention preservation + idempotency) ---------------------
-    $ExistingById = @{}
-    $ExistingLastModified = $null
-    if (Test-Path -LiteralPath $OutPath) {
-        try {
-            $prior = Get-Content -Raw -LiteralPath $OutPath -Encoding utf8 | ConvertFrom-Json
-            if ($prior.PSObject.Properties['last_modified']) { $ExistingLastModified = [string]$prior.last_modified }
-            if ($prior.PSObject.Properties['containers'] -and $prior.containers) {
-                foreach ($p in $prior.containers.PSObject.Properties) { $ExistingById[$p.Name] = $p.Value }
-            }
-        }
-        catch {
-            Write-Verbose "Existing $OutPath unreadable ($($_.Exception.Message)); rebuilding from scratch."
-        }
-    }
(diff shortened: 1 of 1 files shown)


jpsnover:
## TL GV — APPROVED. Un-draft + self-merge on green. This clears the cmdlet half → I release G8a/G8b once it's on main.

Read the full refactor (not just the numbers); it's a clean read-only hoist with the lost-update surface correctly retained. All four guardrails satisfied:

1. **Existing-read stays INSIDE the lock, as the merge basis ✅** — `$ExistingById` is read as the first operation inside the `try` (right after `Enter-GroundingLock`), and the build loop merges the outside-computed candidates against THAT fresh in-lock read. So a concurrent reconciler `node:*` write landing during the ~296s scan is picked up (not a lost update). This is the load-bearing invariant and it holds.
2. **sha unchanged-check vs the in-lock read ✅** — the human-mention-preservation compares `$scan.Sha` (source-derived) against the existing container's `text_sha256` from the in-lock `$ExistingById`.
3. **Output-equivalent ✅** — 2846 own / 4087 mentions unchanged, 36/36 Pester. Structurally sound: the hoist just computes the same nfc/sha/candidates earlier; `$Containers` and `$AliasEntries` are both source-derived (the reconciler never writes SEI/summary source or entities), so collecting them outside the lock yields identical content. Serially (today, G8 off) output is identical; under concurrency the later ExistingById snapshot is the CORRECT fresh basis.
4. **Hold ≤60s ✅** — 296.5s → **1.5s**, well under both the 60s wait and the 120s stale-break. Telemetry (`Write-Verbose` hold time, stopwatch stopped before release) matches the reconciler's per-run log.

Verified independently: (a) key-sort order preserved — `$ScanByCid` is `[ordered]`, built over `$Containers.Keys | Sort-Object`, and the in-lock loop iterates `$ScanByCid.Keys` in that order; (b) no `entity_mentions.json` read remains outside the lock; (c) no `#1806`-style hazard — the cmdlet does a single atomic tmp+rename write with no separate ledger, so there's no write-ordering trap; the `finally` releases on every path incl. the `unchanged` early return.

Symmetry with the reconciler (#1806, on main): both tools now hold the lock ONLY around the `entity_mentions.json` read→merge→write, computing everything read-only outside — 0.01s (reconciler) / 1.5s (cmdlet), both far under 120s. The lock-symmetry sign-off (t/3163#4) is fully satisfied.

Merge on green (base=main, head==d5ed70e1, CI green on that OID). On this landing on main, both grounding-writer holds are safe → I release G8a (t/3171 GROUNDING_RECONCILE_INLINE) + G8b (t/3172 GROUNDING_SWEEP_ENABLED). Excellent, disciplined work — measured over the real corpus, found the real cost, fixed it at the right granularity.

PR #1812, opened 2026-09-02
refine(eslint): degraded-return predicate → []/{}/null/undefined only (t/3200 GV) [GATED: TL GV]

jpsnover:
Refines `local/require-warn-on-degraded-catch-return` per TL's GV predicate (p/342) — TL approved + expedited. Fixes a drift from t/3200#2: the shipped rule flagged ALL `Literal`/`Object` returns, so it also flagged boolean guards (`return false`) and observable Result objects (`{ok:false,reason}`), inflating the live t/3219–t/3223 cleanup by ~half.

**Predicate now:** FLAG catch→ `[]` / `{}` / `null` / `undefined` (silent no-data / ADR-001 graceful-empty). EXCLUDE boolean/number/string guards AND non-empty/observable objects & arrays (failure visible in the shape → not silent). Bare `return;` excluded (indistinguishable from void control-flow w/o type info). Rationale co-located in the rule.

**Test:** both-arms RuleTester **20/20** — FIRES on `return []`/`null`/`{}`/`undefined` (IO, no recorder); SILENT on `return false`, `return {ok:false,reason}`, non-empty arrays, and any catch that logs first.

**Measured on post-cleanup main (t/3219–t/3223 landed):** taxonomy-editor/src drops from 37 → **1 genuine site** (`src/main/ipc/sourceHandlers.ts:99`, a silent `return undefined` on an fs-read — ElectronMain's t/3221 has only the base-rule marker, not this rule's satisfier). Also surfaces **4 orphaned unused-disable** warnings (resilience.ts, analyticsEmitter.ts, featureFlags.ts ×2 — owners annotated FPs before TL's hold; now unnecessary). `eslint src/` exits 0 (unused-disables are warnings, under the 4256 cap) — CI-safe.

**Residuals for the t/3205 flip (separate):** (a) ElectronMain fixes sourceHandlers.ts:99 → 0 rule flags; (b) owners drop the 4 orphaned FP disables. Then flip warn→error → TL GV.

Gate-touching → draft; Main TL GV (you designed the gate). Advances t/3200/t/3205.

### DIFF
--- a/lib/eslint-rules/require-warn-on-degraded-catch-return.js
+++ b/lib/eslint-rules/require-warn-on-degraded-catch-return.js
@@ -1,10 +1,13 @@
 // Custom ESLint rule (t/3200, Fallback-Path Logging — follow-up to t/3169).
 //
-// A `catch` that returns a DEGRADED default (a `Literal` / `[]` / `{}`) FROM A DATA/COMPUTE/IO PATH
-// is a fallback by definition (the t/3165 genus: cache-miss→recompute, primary→secondary,
-// retry-exhausted→default, ADR-001 graceful-empty), so it must record at WARN or ERROR stating what
-// fell back and why. Info-level and the `/* silent by design */` escape-hatch do NOT satisfy it —
-// those were the two gaps the base `require-flight-recorder-in-catch` rule left open (t/3169#2).
+// A `catch` that returns a SILENT empty/no-data default (`[]` / `{}` / `null` / `undefined`) FROM A
+// DATA/COMPUTE/IO PATH is a fallback by definition (the t/3165 genus: cache-miss→recompute,
+// primary→secondary, retry-exhausted→default, ADR-001 graceful-empty), so it must record at WARN or
+// ERROR stating what fell back and why. Info-level and the `/* silent by design */` escape-hatch do
+// NOT satisfy it — those were the two gaps the base `require-flight-recorder-in-catch` rule left open
+// (t/3169#2). The predicate is deliberately NARROW (TL t/3200 GV, p/342): boolean/number/string guard
+// returns (`return false/0/''`) and NON-EMPTY / observable objects (`{ ok:false, reason }`, `{ error }`)
+// are EXCLUDED — their failure is visible in the return shape, so it isn't silent. See isSilentEmptyReturn.
 //
 // This is a SEPARATE rule from the base one on purpose: ESLint severity is per-rule-id, so keeping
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL GV — APPROVED. Matches the t/3200 predicate spec exactly; both arms proven. Merge on green.

`isSilentEmptyReturn` is precisely the p/342 spec:
- **FLAG** (silent no-data): empty `[]` (`elements.length===0`), empty `{}` (`properties.length===0`), `null` (`Literal value===null`), explicit `return undefined` (`Identifier name==='undefined'`). Each is indistinguishable from success-with-no-data → the ADR-001 graceful-empty / t/3165 class.
- **EXCLUDE** (not silent): boolean/number/string literals (`return false/0/''` → `value===null` is false → not flagged) and NON-EMPTY / observable objects & arrays (`{ok:false,reason}`, `{error}`, `[x]` → length>0 → not flagged) — the failure is visible in the return shape.
- Bare `return;` deliberately not flagged (no type info → indistinguishable from void early-exit; correct over-flag guard).

**Both arms (RuleTester 20/20):** VALID asserts `false` / `0` / `{ok:false,reason}` / `{error}` / `[fallbackItem]` are NOT flagged **even without a recorder** (the exclusion arm); INVALID asserts `{}`+info-level, `null`, and `undefined` on an IO path ARE flagged (the fire arm, incl. info-insufficient). Stays at `warn` — no premature error-flip (that's t/3205, separately GV-gated). CI-safe.

Small nit (optional, non-blocking): the empty-`[]`-with-no-recorder FIRE case isn't dedicated-tested (the `{}` fire uses the info-level case; null/undefined have no-recorder cases). The array-empty branch is symmetric to the object-empty branch that IS tested, and the non-empty-array exclusion IS tested, so coverage is adequate — add one `return []`-no-recorder invalid case if you want full symmetry.

Approved — merge on green (head 83084f0a == ls-remote tip, verified; base=main; CI green on that OID).

**Downstream to the t/3205 error-flip (37→0):** after this lands (drops taxonomy 37→1), the flip precondition needs (a) ElectronMain's 1 genuine fix (`sourceHandlers.ts:99`) and (b) dropping the 4 orphaned FP `eslint-disable`s — those are exactly the FP-churn my p/342 hold was meant to prevent (owners annotated before the hold reached them), so removing them is correct cleanup, not a regression. On 37→0, route the t/3205 error-flip to me for GV (Gate Promotion: confirm zero-tolerated-noise + both arms before flipping warn→error).

PR #1823, opened 2026-09-02
fix(cl): strip attitude-attribution clause in logical_form prompt (t/3227)

jpsnover:
## What
Prompt hardening from the D3b finding (t/3126): the residual `predicate`=0.50 is largely **attitude leaking into the predicate** on meta-descriptive BDI claims. b-04 emitted `support(agent:skeptics,...)` (duplicating `modality.desire`) instead of the embedded `protect(GDPR, individual-autonomy)`; b-13 emitted the surface `promote(...)` instead of the deeper belief `prevent(monopolization)`.

The existing meta-descriptive rule strips the *reporting* frame ("The document discusses…") but not the *attitude-attribution* clause ("Skeptics support…", "reflects the belief that…").

## Fix
Adds one rule: the stance is already carried in `modality` (holder+attitude), so the predicate must be the CONTENT proposition, never the stance verb. Distinguishes stance verbs (support/oppose/believe/want/aim/seek/intend/"view…as") from content action verbs (remove/lower/encourage/protect/prevent/consolidate), with a worked GDPR example. `/review-prose`: 0 new em-dashes.

## Why draft (gated)
This changes formalization **behavior**, so it must be validated before landing — I will not merge blind:
1. **Validation probe** (PowerShell, dry, no data write): re-run ~8 claims with the new prompt — the 4 meta-descriptive predicate-misses (b-04/b-09/b-13/b-23) for lift + 4 currently-correct (b-05/b-07/b-19/b-28) for no-regression — scored vs `golden_set_d3b.json`.
2. **CL Main** mandatory review (prompt = CL-owned formalization-behavior surface).

Un-draft only when both clear. Expected: b-04 support→protect, b-13 promote→prevent; b-05/b-07/b-19/b-28 unchanged. Validation can piggyback on the t/3229 top-up run.


jpsnover:
## CL Main mandatory review — PR #1823 (logical-form attitude-attribution rule, t/3227)

**Verdict: approve-with-notes.** The rule is ontologically correct, well-scoped, and correctly gated. Un-draft + merge on the PS validation probe (the binding empirical gate) coming back green; my review gate is cleared.

**Why it's right.** A stance verb as `predicate` double-counts the attitude that reification already puts in `modality` (holder + attitude). Emitting `support(agent:skeptics)` when `modality.desire` already encodes the skeptic's stance is exactly the duplication the neo-Davidsonian design avoids — the predicate must carry the CONTENT proposition (`protect(GDPR, individual-autonomy)`). The stance-verb vs content-action-verb distinction (support/aim/seek vs protect/remove/lower) is the correct operational test, and the GDPR worked example demonstrates it together with the meta-descriptive frame-strip. Scope discipline is good: attribution clause only, `about[]` untouched (correct post-#1820).

**Notes (non-blocking):**
1. **Stance-verb list — minor completeness gap.** support/oppose/believe/think/hold/want/desire/aim/seek/intend/"view…as" covers the common cases, but a few frequent AI-policy stance verbs are absent: **advocate, endorse, favor, reject, call for, push for, prioritize** (and "warn (that)"). The rule states the *principle* (stance duplicates modality) so the model should generalize, but adding 3–4 of these hardens it against the long tail. Optional.
2. **"view … as" edge.** Guard that it doesn't over-strip a genuinely definitional/copular content claim (e.g. "the report classifies X as a high-risk system" where the classification IS the assertion). Low risk given the principle framing; worth a watch in the probe output.
3. **Provenance.** A prompt-rule change to a metric input (feeds the D3b `predicate` score) is a provenance event — on merge, note the prompt-version bump + the D3b re-baseline once the probe confirms the lift, so the register/golden trail reflects it.
4. **Empirical gate is the binding one.** My approval is on formalization correctness; the predicate lift (b-04/b-13) + no-regression (b-05/b-07/b-19/b-28) is confirmed by your PS probe vs `golden_set_d3b.json`. Correctly holding un-draft on it. The honest scoping of b-09 (synonymy) / b-23 (clause selection) to t/3228 is the right call — don't stretch this rule to cover them.

Good, tightly-scoped fix. Clear my gate; land on the probe.

PR #1828, opened 2026-09-02
obs(server): surface embedding offload-flag + resolved chain to stdout (t/3165)

jpsnover:
## t/3165 observability — the self-diagnosing marker → **Main (TL) GV** (approved e/136#10)

The prod regression (offload "looked engaged at the durable #1793 flip but silently ran in-process ONNX on the main thread") was diagnosable only by inference: `computeEmbeddings` records its resolved `chainMembers` to the **FR ring only** (aiBackends:821), invisible to Log Analytics. Root cause (traced): the chain is `if(flag)→onnx-batch-worker / else-if→onnx-batch(in-process)`; main-thread ONNX + zero worker logs + no :827 error is **only** producible by the else branch ⇒ `isEmbeddingWorkerOffloadEnabled()` evaluating **false at runtime** (likely the bicep value-form — DevOps `printenv` confirming).

### Change (observability-only, no behavior change)
- **aiBackends.`computeEmbeddings`** — one Pino/stdout INFO at chain resolution: `{ offloadFlag, chainMembers, requester, inputCount, cacheHits, cacheMisses }`. Unconditional + pre-compute, so it can't be masked by a swallowed downstream error. Next prod debate self-reports `offloadFlag=false + chain=['onnx-batch']` (the starvation shape) vs `true + ['onnx-batch-worker']` (engaged).
- **server.ts boot** — one line logging the offload flag's runtime value, so a deploy's effective `EMBEDDING_WORKER_OFFLOAD` is greppable at startup (surfaces the "looked set in ACA config, evaluated false" gap directly).

### Verify
`tsc -p tsconfig.server.json` clean; eslint 0 errors. No test — pure logging.

Note: if `printenv` confirms the flag is false, the actual FIX is a bicep/env value-form correction (DevOps), not code — this PR is the durable self-diagnosis either way. Draft pending your GV.

### DIFF
--- a/taxonomy-editor/src/server/ai/aiBackends.ts
+++ b/taxonomy-editor/src/server/ai/aiBackends.ts
@@ -814,6 +814,20 @@ export async function computeEmbeddings(
     chain.push({ name: 'onnx-batch', compute: (t) => onnxComputeEmbeddings(t) });
   }
 
+  // t/3165 observability: emit the RESOLVED chain + the offload-flag value to Pino/stdout. The
+  // completion record at :821 lands in the FR ring ONLY (invisible to Log Analytics) — which is
+  // precisely why the "offload looked engaged at flip but silently wasn't" regression took inference
+  // to diagnose. A prod compute now self-reports on stdout: offloadFlag=false + chain=['onnx-batch']
+  // (in-process, main-thread — the t/3165 starvation shape) vs offloadFlag=true +
+  // chain=['onnx-batch-worker'] (engaged). Unconditional + pre-compute → can't be masked by a
+  // swallowed downstream error.
+  log.api.info({
+    component: 'ai-backends', requester,
+    offloadFlag: isEmbeddingWorkerOffloadEnabled(),
+    chainMembers: chain.map(c => c.name),
+    inputCount: texts.length, cacheHits, cacheMisses,
+  }, 'computeEmbeddings chain resolved');
+
   try {
     // t/2985: timeout is now per-chunk inside resolveEmbeddingsChunked — no aggregate ceiling.
     const result = await resolveEmbeddingsChunked(texts, ids, local, chain);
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL GV — APPROVED. Merge on green. Pure observability, exactly the approved design.

Both lines are log-only (no behavior change), and the placement is right:
- **aiBackends `computeEmbeddings chain resolved`** — emitted AFTER chain construction, BEFORE the `try`/compute → unconditional + pre-compute, so a swallowed downstream error can't hide it. Carries `offloadFlag` + `chainMembers` to Pino/stdout (Log Analytics), so the next prod debate self-reports `offloadFlag=false,chain=['onnx-batch']` (the starvation shape) vs `true,['onnx-batch-worker']` (engaged). The completion record only ever hit the FR ring (:821) — invisible to Log Analytics — which is exactly why this regression took inference to find. Bonus: `cacheHits/cacheMisses` on the same line also lights up the cache-hit load-profile question.
- **server.ts boot log** — one-line runtime flag value at startup → a deploy's *effective* EMBEDDING_WORKER_OFFLOAD is greppable at boot, so "set in ACA config but process.env evaluated false" is immediate, not inferred.

Real tip == `git ls-remote` (6260bc18) — no stale-head risk. tsc+eslint clean. Merge on green (base=main, head==6260bc18, CI green on that OID).

This is the durable closer for the core t/3165 observability gap: it would have surfaced a wired-but-ineffective flag at the flip instead of a full incident later, and it confirms the root cause on the NEXT debate without needing the interactive printenv (which stays the faster immediate check).

Non-blocking note: the per-compute INFO line fires on every computeEmbeddings — fine now (we WANT the visibility for the incident + steady-state flag confirmation via the boot line). If it proves chatty post-incident, sample it or drop to debug and rely on the boot line. Don't gate on it.

PR #1832, opened 2026-09-02
fix(server): serialize-once cache for synthetic-embeddings — kills the 3s main-thread block (t/3165)

jpsnover:
## t/3165 — the incident fix → **Main (TL) GV** (design LOCKED e/136#37)

**Root cause (converged):** the offload works (worker off-main, `printenv=1`, cache resolves). The prod 3018ms event-loop block was **not** the debate `/compute` serialize (all batches ≤512 ≈ 350ms) — it mapped 1:1 to **`/api/taxonomy/synthetic-embeddings`**: a STATIC ~4144-vector / ~400MB corpus that `loadSyntheticEmbeddings()` **rebuilds every call** (per-POV `.npy` network read + `parseNpy` + build) and the route re-`JSON.stringify`'d un-yielded on main — **once per debate** (`taxonomyContext` fetches the full corpus). ~400MB heap spike + ~3s block.

### Fix (route-level, my scope)
- **Serialize once → cache the Buffer**; serve near-free on every later GET (skips load/parse/build/serialize). **Promise-dedupe** (cache the in-flight promise) so a cold-start burst launches ONE build (= DevOps's single-flight, subsumed — no separate guard).
- **`httpKit.jsonStringifyChunked`** — byte-identical to `JSON.stringify`, yields the loop every N top-level items → even the COLD serialize can't block past the liveness deadline.
- **Cold-path phase-timing** (`load_ms` / `serialize_ms` / heap before+after) → the redeploy self-proves: next real synth GET shows `serialize_ms`→~0 (cache hit) + heap flat + no block.
- **`__invalidateSyntheticEmbeddingsCache()`** exported + a co-located **write-frozen contract** comment (process-lifetime cache is safe only while the corpus doesn't mutate; no server synth-mutation route exists today — if one is added it MUST invalidate).

Response body **byte-identical** → zero client change. **Anon auth-gate NOT included** (TL call: t/3230 already blocks the debate trigger for anon; anon-access + the payload-subset redesign ride TL's fast-follow).

### ⚠ Deviation (flagged for GV)
Fix set listed a `/compute` cap + chunk-yield as belt-and-suspenders. **Deferred** — Diagnostics disproved `/compute` as the culprit (all ≤512 ≈ 350ms), so a never-firing cap + a yield on a non-culprit path would add blast radius (2 more files) to an incident PR for zero incident value; the `jsonStringifyChunked` helper is in place for a trivial follow-up. Say the word in GV and I'll add it.

### Verify
`tsc -p tsconfig.server.json` clean; eslint 0; **`syntheticEmbeddingsCache.test.ts` 16/16** (byte-identity across array/object/null/edge cases + cache/dedupe/invalidate/retry-on-failure); routeTable 4/4 (registration unchanged).

### Close criteria (post one-bundle redeploy)
On a REAL synth GET: `serialize_ms`→~0 (cache hit) + heap bounded + no event-loop block; Diagnostics co-verifies calm loop. Prevention (loop-sampler covers the serialize phase, tie t/3210) filed separately.

### DIFF
--- a/taxonomy-editor/src/server/__tests__/syntheticEmbeddingsCache.test.ts
+++ b/taxonomy-editor/src/server/__tests__/syntheticEmbeddingsCache.test.ts
@@ -1,0 +1,105 @@
+// @vitest-environment node
+// Copyright (c) 2026 Jeffrey Snover. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root.
+
+// t/3165 — the synthetic-embeddings serialize-once cache + the httpKit chunk-yield serializer.
+// Root cause: /api/taxonomy/synthetic-embeddings rebuilt (network .npy read + parse + build) AND
+// re-`JSON.stringify`'d the STATIC ~4144-vector / ~400MB corpus on every debate-triggered GET — an
+// un-yielded ~3s main-thread block + ~400MB GC churn. Fix: serialize once, cache the Buffer, serve it
+// near-free; promise-dedupe so a cold-start burst launches ONE build; chunk-yield the cold serialize.
+//
+// These lock: (1) jsonStringifyChunked is byte-identical to JSON.stringify (arrays/objects/null/edge
+// cases) so the response contract is unchanged; (2) the cache serves from memory (loader NOT re-called),
+// dedupes concurrent cold callers to one build, and the invalidator forces a rebuild.
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const { loadSynth } = vi.hoisted(() => ({ loadSynth: vi.fn() }));
+vi.mock('../storage/fileIO.js', () => ({ loadSyntheticEmbeddings: loadSynth }));
+vi.mock('../../../../lib/flight-recorder/index.js', () => ({ getGlobalRecorder: () => ({ record: vi.fn() }) }));
+
+import { jsonStringifyChunked } from '../httpKit.js';
+import { __getSyntheticEmbeddingsBufferForTest as getBuffer, __invalidateSyntheticEmbeddingsCache as invalidate } from '../routes/taxonomy.js';
+
+describe('jsonStringifyChunked — byte-identical to JSON.stringify (t/3165)', () => {
+  // yieldEvery=2 forces the yield path even on tiny inputs.
+  const cases: Array<[string, unknown]> = [
+    ['number[][] array', [[0.1, 0.2, 0.3], [0.4, 0.5, 0.6], [0.7, 0.8, 0.9]]],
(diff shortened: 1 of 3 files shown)


jpsnover:
## TL GV — APPROVED. Merge on green. Two minor, non-blocking notes.

Verified against the #37 checklist — the incident fix is correct and well-tested:

- **Promise-dedupe (single-flight) ✓** — warm hit returns the cached Buffer; a concurrent cold burst all await `_synthEmbeddingsInFlight` = ONE build (tested); the `finally` clears the in-flight handle so a FAILED cold build is not cached and the next call retries (tested). Correct.
- **`jsonStringifyChunked` byte-identical ✓** — only the TOP level is chunked; every element/value + key routes through real `JSON.stringify`, so escaping (`v"quote`), nested `toJSON`, key order, undefined/function/symbol omission, and array-hole→`null` all match. Empty `[]`/`{}` and the primitive/null fallback are handled. The 10 edge-cases + 500-entry round-trip lock it. 16/16.
- **Invalidator + write-frozen contract, co-located at the cache site ✓** (Gate Co-Location, exemplary). Phase-timing on the cold path only (`load_ms`/`serialize_ms`/heap) — warm hit is silent = the self-proof. Response byte-identical (`sendJsonBuffer`, idempotent-write guard) → zero client change.
- **Memory:** the cold path's ~400MB transient object is GC'd after serialize; only the (~tens-of-MB text) Buffer is retained → net win vs the repeated per-call 400MB alloc.
- **Deviation accepted:** /compute belt-cap deferred (Diagnostics disproved /compute; the httpKit helper is in place if ever needed). Fine.

### Minor notes (non-blocking — no trigger today; do now or document)
1. **Invalidator + in-flight race (latent, future).** `__invalidateSyntheticEmbeddingsCache()` clears `_synthEmbeddingsBuffer` but not an in-flight build. If a synth-mutation route is ever wired and calls invalidate() DURING a concurrent cold build, that in-flight build (of pre-write data) still resolves and caches stale bytes. No mutation route exists today, so inert — but extend the write-frozen contract comment to note: a future mutator needs a generation-guard (bump a counter on invalidate; the IIFE checks its generation before caching), not just a buffer-null. Cheap to add the note now so the next author doesn't miss it.
2. **`jsonStringifyChunked` top-level `toJSON` (generic-helper caveat).** A top-level object with a `.toJSON()` would be iterated by-entries instead of honoring toJSON (JSON.stringify would call it). Non-issue for the plain synthetic-embeddings Record, and the doc says "plain object" — just add one line that the helper assumes a plain array/object at the top level (no top-level toJSON), so a future caller on a class instance doesn't get surprised.

Neither blocks. Un-draft + self-merge on green (base=main, head==55e305a2 == ls-remote, CI green on that OID). Then it rides the one bundled redeploy. On the flagged scope call in the PR — reply here and I'll rule. Excellent, precise fix — the cold-path phase-timing is exactly the self-proving instrumentation the close criteria need.

PR #2006, opened 2026-09-05
feat(qbaf): fork-B census-merge applier + frozen verified set (t/3339) — TL GV

jpsnover:
**The single corpus write for fork-B** (benefit arm cleared: union ×2.82 / 11.6%, t/3339#14). Routing to TL for GV before the owner runs `--write`.

**TL's 3 conditions (t/3339#329) — all met:**
1. **Explicit committed verified set, no re-derive:** `scripts/fork-b-verified-merges.json` freezes 62 same-doc contradict@≥0.90 (golden #1942) + 12 cross-conflict PI-dual-verified (16→≤1/stance). The applier reads only these IDs — no re-classify, no auto-merge path.
2. **Provenance tags:** every edge/new-conflict carries `claim_origin=observed`, `edge_origin=semantic`, `symmetric=true`, `classifier`+`tau`+`confidence`, `register=t/3342`.
3. **0-collateral verifier:** dry-run diffs every untouched conflict byte-identical before any push.

**Dry-run on the live corpus (verification artifact):**
- 20 same-doc conflicts touched (+124 symmetric edges), 12 new cross-conflict conflicts, 1240→1252.
- **COLLATERAL: ZERO** (every untouched conflict byte-identical).
- Merged corpus reconciles **exactly** to 48 conflicts-with-attacks (×2.82) / 120/1034 nodes (11.6%).
- DF-QuAD strengths recomputed via bridge (2-cycles convergent, t/3339#12). 3 pytest green.

**Execution split:** this PR is code only — no data written. After GV, the **owner** runs `python scripts/apply_semantic_merge.py --write` against ../ai-triad-data (PS never writes the data repo directly); `--write` refuses unless verified-clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2010, opened 2026-09-05
fix(qbaf): same-subject hard-gate on the numeric detector (t/3337)

jpsnover:
Standalone correctness fix (CL spec t/3337#5): the numeric detector was **0/54 precision** on the same-doc census — the ≥3-shared-word subject proxy fired on different-subject pairs ('17% postings' vs '24% skills'). A 0%-precision detector would corrupt any corpus write, so this is worth landing on its own; the same-doc *lever* revival stays separately gated on TL's recalibrate.

**Fix (before firing):** number-mask both assertions → **masked-cosine (MiniLM) ≥ 0.93** + same unit-type. Candidate check runs first, so embed is skipped when there's no disjoint-number candidate.

**Calibration:** fires 6/6 constructed same-subject, blocks 52/54 production FPs (CL). Real-MiniLM check: same-subject cos 1.00 (fire); automation postings-vs-skills 0.71 (skip); latency-vs-cost 0.76 (skip).

**Accepted v1 residual (CL):** same-metric-different-scope + 63/65 rounding pass the subject gate (masked-cosine can't separate; LLM arbitrates) — value-floor add-on deferred (not v1).

Tests +6 → **35 pytest green**. Embedder injectable for determinism.

Flow: CL re-certifies on the same-doc census → TL GV.

### DIFF
--- a/scripts/enrich_conflicts_qbaf.py
+++ b/scripts/enrich_conflicts_qbaf.py
@@ -245,27 +245,58 @@ def _cc_bare_numbers(text):
     return out
 
 
-def _detect_numeric_temporal_conflict(text_a, text_b):
-    """True iff A and B share a subject AND cite directly incompatible quantities of the same kind.
+_CC_MAG_RE = re.compile(r"\b(billion|million|trillion|thousand|fold)\b", re.IGNORECASE)
+_CC_CUR_RE = re.compile(r"[$€£]")
 
-    Conservative (high precision): requires >=3 shared content words, and a same-kind numeric set that
-    is non-empty on BOTH sides and disjoint (percentages, years, or bare numbers). Bare numbers demand
-    a stronger subject overlap (>=4) since they are the most ambiguous. Returns False on any doubt.
+
+def _mask_numbers(text):
+    """Replace numeric values (%, years, currency, magnitudes, bare numbers) with a NUM placeholder so an
+    embedding of the result captures the METRIC/subject, not the values (t/3337#5 same-subject gate)."""
+    t = text or ""
+    t = _CC_PCT_RE.sub(" NUM ", t)                       # percentages / 'percent' / pp
+    t = _CC_YEAR_RE.sub(" NUM ", t)                      # years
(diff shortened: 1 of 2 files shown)


jpsnover:
## TL GV — APPROVED (merge on green). Closes t/3337.

Verified the numeric-detector same-subject fix against the t/3339#20 #3 conditions:

**Predicate is sound.** Fires only on: same numeric-kind + disjoint values + masked-cosine ≥0.93 (masked text = subject, so ≥0.93 ⇒ same metric). This directly kills the 0/54 FP mode — 'X% fewer postings' vs 'Y% fewer skills' mask to different subjects → cosine <0.93 → SKIP. Same-metric rounding (63% vs 65%) masks-identical → passes the subject gate → correctly deferred to the LLM rather than auto-fired. That's the honest boundary.

**Both arms tested (the load-bearing requirement).** `test_..._blocks_different_subject_that_old_detector_fired` (fire-suppression: the automation-study FP now returns False) + `test_..._still_fires_same_metric` (recall: same-metric-different-value still True). Plus the no-candidate short-circuit and the mask helper. The `_bow_embedder` proxy exercises the gate logic deterministically in CI; the 54→1 census (t/3337#6, real MiniLM) is the real-distribution validation. Logic + distribution both covered.

**Gate Co-Location ✓** — the accepted v1 residual (same-metric-different-scope 'top 10%' vs 'top 20%') is documented in the docstring at point of use, not buried in ticket history. **Threshold basis ✓** — 0.93 is calibrated on the census with the t/3337#5 pointer, not bare-stipulated.

**On the 'precision 0/1' framing:** correct and honest — the detector reads ~0 precision only because this corpus has zero true numeric contradicts to catch; the success metric here is FP-elimination (98% blocked) + preserved recall (constructed positives fire), not precision on an empty positive class. The detector is correctly quiet, not dead.

This is a correct standalone correctness fix that protects **any** future write using numeric contradicts, independent of the retired same-doc bar. 

**Merge:** single-scope, not joint-gv → self-merge on green CI with `--match-head-commit` (CI currently in_progress; do not merge red — the #2006 lesson). t/3337 closes on merge.

PR #2011, opened 2026-09-05
feat(devops): Phase-1 orphan detection in staging env sync (t/3345)

jpsnover:
## Summary
- Adds `-NamesOnly` switch to `Get-BicepBaseEnv.ps1` — captures all key names bicep declares, including non-literal values (interpolated, resource-ref, param-ref) that the standard literal-value pass skips
- `Sync-StagingEnv.ps1` now detects orphaned env keys (present in live ACA template but absent from bicep managed set) and emits `::warning::` with manual-removal command
- Fail-closed guard aborts if NamesOnly count < literal count to prevent mass-wipe on parse failure
- `az containerapp update` only fires when drift exists; orphans are warn-only (Phase-2 pending TL GV)

## Test plan
- [x] `GetBicepBaseEnv.Tests.ps1` — 5 new tests: NamesOnly superset, non-literal coverage, string array type, no duplicates
- [x] `StagingEnvSync.Tests.ps1` — +4 tests: orphan DryRun exits 2, ALLOWED_ORIGINS (non-literal) not flagged as orphan, secretRef not flagged, existing pass/fire arms unchanged
- [x] All 10/10 pass locally

## Hold
Draft until TL sign-off per gate-promotion discipline. Phase-2 (`--remove-env-vars`) in a separate PR after >=1 staging cycle observation.

### DIFF
--- a/operations/devops/Get-BicepBaseEnv.ps1
+++ b/operations/devops/Get-BicepBaseEnv.ps1
@@ -39,7 +39,8 @@
 [CmdletBinding()]
 param(
     [string]$BicepPath,
-    [switch]$ForStaging
+    [switch]$ForStaging,
+    [switch]$NamesOnly
 )
 
 Set-StrictMode -Version Latest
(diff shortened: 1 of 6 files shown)


jpsnover:
HOLD (gate-promotion discipline): Draft until TL sign-off on Phase-1 GV. Per t/3345 approved design, Phase-2 (live --remove-env-vars) in a separate PR after >=1 staging cycle observation.

jpsnover:
## TL GV — Phase-1 APPROVED with ONE required fix + ONE recorded Phase-2 blocker

Solid Phase-1. Correct against most of t/3345#2:
- **Warn-only ✓** — `--remove-env-vars` appears only inside the `::warning::` as a *manual suggestion*, never executed; Phase-2 live removal explicitly deferred. Real-env-first honored.
- **Fail-closed on gross collapse ✓** — null / empty / count-below-literal → `Write-Error` + exit 1.
- **Safety tests ✓** — ALLOWED_ORIGINS (non-literal) in ManagedNames → not orphaned; secretRef excluded from CurrentMap; DryRun exit-2 on orphan.

**① REQUIRED before Phase-1 merge — completeness test must hit the REAL main.bicep (condition #3).** Every completeness test runs against `fixtures/bicep-env-drift/good-main.bicep`, not `deploy/azure/main.bicep`. This is the crux of condition #3: a fixture is authored to match your regex, so it *structurally cannot* catch a `name:` format that exists in the real bicep but the regex misses — the exact under-matching failure that produces a false orphan (→ a misleading warning now, a live deletion in Phase-2). Add one test that runs BOTH `-NamesOnly` and hashtable-mode against the real `deploy/azure/main.bicep` and asserts: (a) `-NamesOnly` ⊇ hashtable keys on the real file, AND (b) it includes the known non-literals (ALLOWED_ORIGINS, AZURE_KEYVAULT_URL, APPLICATIONINSIGHTS_*). Cheap (the file's in-repo) and it's the assertion that makes the whole switch trustworthy. Un-draft + merge once this is added and green.

**② RECORDED Phase-2 blocker — strengthen the runtime guard from count to membership-superset before enabling live removal.** The guard is `$ManagedNames.Count -lt $BicepEnv.Count`. In Phase-1 (warn-only) that's fine — a dropped literal key → false *warning*, no harm. But before Phase-2's live `--remove-env-vars`, a parse bug that drops literal key K while count stays ≥ (non-literal pass picks up a spurious name) would flag K as orphan → **delete a live managed key**. A count check can't catch that; membership can. Replace with: `foreach ($k in $BicepEnv.Keys) { if ($k -notin $ManagedNames) { Write-Error; exit 1 } }` — the true fail-closed invariant. Not required for this PR; it's a hard gate on the Phase-2 GV.

①  closes Phase-1; ② is on the Phase-2 checklist. Good, careful work — the warn-only staging + explicit Phase-2 defer is exactly the real-env-first shape I asked for.

PR #2052, opened 2026-09-07
feat(validation): aggregate strip-count WARN + logical_form frame-count CI floor (t/3378)

jpsnover:
## Summary
Observability half of the t/3375 prevention-per-incident follow-up (t/3376 is the write-side fix, separate PR).

- **Correction to the ticket's premise**: the load-time strip WARN already existed (t/3250) — this PR adds the missing aggregate: a per-POV strip-count summary, escalating `warn`->`error` above 5 frames stripped in one load.
- **Frame-count CI floor** in `validation.data.test.ts`: schema-parse passes even when `logical_form` (optional) silently vanishes — this asserts each POV file carries at least its baseline frame count, catching a systemic drop like t/3375's 133-frame loss that schema validation structurally cannot.
- Baselines verified against live data: acc 133, saf 274, skp 234 (641 total) — exact-floor per TL's "intended friction" call (t/3375#2).

## Test plan (GV both-arms evidence)
- **Strip summary** (`useTaxonomyStore.test.ts`, 3 new tests): clean load emits no summary; a load with strips emits one aggregate WARN per POV with the correct count; count > 5 escalates to `level: 'error'`. All passing.
- **Frame-count floor** (`validation.data.test.ts`): manually verified an inflated floor value (9999) fails loudly with the live count and floor in the assertion message, then reverted to the real baseline (133/274/234) which passes. Not committed as a test (would require mocking the data file) — verified by hand per the design's "both-arms" requirement; live baseline arm is in the suite.
- `npx vitest run` on both changed test files: 137/137 passing.
- `npx tsc --noEmit -p tsconfig.json`: clean.
- `npx eslint` on all 3 changed files: 0 errors (4 pre-existing complexity warnings in `taxonomyDataSlice.ts`, unrelated to this change).

Refs: t/3378, t/3375 (incident), t/3376 (write-side fix, separate PR), t/3250 (original per-node WARN)

### DIFF
--- a/taxonomy-editor/src/renderer/hooks/useTaxonomyStore.test.ts
+++ b/taxonomy-editor/src/renderer/hooks/useTaxonomyStore.test.ts
@@ -56,9 +56,15 @@ vi.mock('@lib/debate/validateNodeId', () => ({
   validatePovNodeId: vi.fn().mockReturnValue({ valid: true }),
 }));
 
+const { mockRecord, mockStripInvalidLogicalForm } = vi.hoisted(() => ({
+  mockRecord: vi.fn(),
+  // t/3250: loadAll calls this per node. t/3378: tests override this per-case to simulate strips.
+  mockStripInvalidLogicalForm: vi.fn().mockReturnValue({ removed: false }),
+}));
+
 vi.mock('@lib/flight-recorder/index', () => ({
   getGlobalRecorder: vi.fn().mockReturnValue({
-    record: vi.fn(),
+    record: mockRecord,
   }),
 }));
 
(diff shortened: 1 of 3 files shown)


jpsnover:
## TL review — APPROVED, merge on green.

**The CI frame-count floor is the right design** — the piece I was most ready to reject as brittle is exactly right: a FLOOR (≥133/274/234), not an exact count, so legitimate additions never red and removals require a deliberate floor edit — with that friction stated as intended, co-located, citing the ruling. It catches what schema validation structurally cannot (logical_form is optional — the t/3375 escape), and the historical fire-arm evidence is real: the incident data (508) would have tripped it. For a one-line ≥ comparison, inspection + the historical case suffice — no seeded arm demanded (proportionality; there's no predicate logic to get wrong).

**Renderer strip summary: clean** — aggregate WARN with count, silent on clean load, warn→error escalation above the co-located systemic threshold (5, with rationale), all three arms tested via the hoisted strip mock.

**Process note (t/3361):** no Second Opinion needed — no schema/data-model change; this extends the existing SO-reviewed safety-net pattern (e/142) with a data-integrity assertion inside an already-required check. Stating that explicitly so nobody re-litigates it.

One future-proofing note, not blocking: when the t/3379 systemic decision lands (§109 amendment or layer regeneration), the floors may need a deliberate update — that's the designed friction working, not a defect.

PR #2053, opened 2026-09-07
fix(taxonomy-data): reattach stripped logical_form before whole-file save (t/3376)

jpsnover:
## Summary
Root fix for t/3375 (133 accelerationist `logical_form` frames silently deleted by a strip-then-save). `stripInvalidLogicalForm` (t/3250) mutates the node object living in store state; a subsequent whole-file `save()` persisted that in-memory deletion back to disk. `save()` now re-reads the on-disk POV file immediately before writing and reattaches any `logical_form` present on disk but missing in memory.

Per TL's two review conditions (t/3376#2), the reattach site's doc comment states:
1. **Load-bearing assumption**: no renderer edit path for `logical_form` exists today (t/3378 confirmed it's unconsumed) — a future edit path must replace this reattach with an explicit tombstone, not blind reattach.
2. **Generalization seam + TOCTOU residual**: this is the one choke point for future strip-at-load fields to reattach at; protection is only as strong as the re-read immediately before write (no lock against a concurrent writer), and whole-file save semantics (last-writer-wins) are otherwise unchanged.

Rejected alternatives (from t/3376#1, TL-approved): strip-for-render-only (bigger refactor, no other field has this split) and a loud-confirm block (no human in the loop for automated saves like harvest-on-save, which is exactly what caused t/3375).

## Test plan
- Both-arms (`useTaxonomyStore.test.ts`, 2 new tests): a `logical_form` present on disk but missing in memory is reattached before save; a node with a valid in-memory `logical_form` is left untouched even when disk carries a different value (proves the reattach only fires on `undefined`, never overwrites a real value).
- `npx vitest run src/renderer/hooks/useTaxonomyStore.test.ts`: 122/122 passing.
- `npx tsc --noEmit -p tsconfig.json`: clean.
- `npx eslint`: 0 errors (4 pre-existing complexity warnings in `taxonomyDataSlice.ts`, unrelated — `save()`'s complexity score is unchanged by this diff).

Refs: t/3376, t/3375 (incident), t/3378 (observability half, separate PR #2052), t/3250 (original strip)


jpsnover:
## TL review — APPROVED with ONE required fix, then merge on green.

Faithful to the approved design: both condition comments in verbatim (load-bearing assumption + tombstone guidance; generalization seam; TOCTOU residual), reattach keyed strictly on `undefined`, count-gated recorder WARN, and the no-reattach arm is nicely discriminating (disk carries a DIFFERENT value, proving memory-present wins).

**Required fix — the silent fail-open at `if (!onDisk?.nodes?.length) return file`:** this skips ALL protection with zero signal, and it's reachable invisibly: on the web profile an ADR-001 graceful-empty read returns exactly this shape — the reattach guard would silently lapse at the very moment a strip could persist. Per the Fallback-Path Logging rule (root AGENTS.md): (1) WARN when the re-read comes back null/empty while the outgoing file has nodes ("reattach protection skipped — on-disk re-read empty"); (2) wrap the `api.loadTaxonomyFile` in try/catch → WARN + return file (fail-open-but-loud) — as written, a THROWN re-read rejects the `.then(save)` chain and kills the user's save, which is the wrong failure direction for a protective pre-step; (3) one test arm: re-read returns null → save still proceeds + the skip-WARN fires.

Small and mechanical — apply, then merge on green with --match-head-commit. The t/3375 write-side hole closes with this PR; #2052 is the detection side (approved separately).

PR #2082, opened 2026-09-07
docs(sage): batch 19 — #206 run-gate execution dead, #207 getByText multi-match

jpsnover:
## Summary

- **#206** [Platform/Safety] ALL Orca `run:` shell-out feedback rules silently inert since late-May 2026 platform regression — `pre-self-merge-verify`, `joint-gv-automerge-guard`, `done-evidence` all dead; static/template rules still fire. Never cite a run-gate as enforcement in a GV without an execution probe; manual disciplines are the only protection until fix. (p/13#43, p/26#106, t/3394, t/3396)
- **#207** [Test/React] `getByText('value')` throws multi-match when the value renders in multiple DOM nodes (args row + `about[]` chip). Fix: `getAllByText(...).length` or `within(container).getByText(...)`. (p/629#1)

## Test plan

- [ ] Verify entries #206–#207 render cleanly and number sequentially after #205

🤖 Sage (Orca) — Observe→Analyze→Record


jpsnover:
ACCURACY skim (t/3362 / Quality TL): #206 and #207 — both accurate. #206: run:/template partial-engine distinction correct; gate inventory (t/3270/t/3318/t/3360) matches; #205/#206 distinction (telemetry vs execution) correctly drawn. #207: getByText strict-single-match semantics correct; getAllByText + within() fixes are valid RTL idioms. Clean — merge on green.

PR #2115, opened 2026-09-09
fix(dwellTracker): heartbeat-credit accrual replaces idempotent open-span model (t/3420 item 2, t/3422)

jpsnover:
## Summary

t/3420 item 2 (TL-approved design, t/3422#1/#2): replaces `EngagementAccumulator`'s idempotent open-span model — the root mechanism behind the idle-tail inflation item 1 patched symptomatically — with per-pulse bounded credits.

- Each `pulse(t)` grants engagement validity through `t + CREDIT_WINDOW_MS` (15s). Pulses within a still-live credit merge into one interval; a gap wider than the window closes the old interval at its expiry and starts fresh — sparse jitter stops purchasing continuous time.
- `pause(t)` always truncates any live interval **at t** (clamped to its credit expiry, whichever is earlier) — the TL-required addition: a pulse-then-hide/subject-change must book only the real elapsed time, never the full window.
- `isEngaged` (unused no-arg getter) → `hasActiveCredit(t)`, matching the class's existing explicit-timestamp convention.
- `onIdleTimeout`/`onHiddenTimeout` simplified to plain `pause(t)` — the accumulator's own credit-expiry clamp now provides the bound, making the old `IDLE_GRACE_MS`-based `Math.min(...)` at those call sites redundant (left in config for wire-compat, doc-noted as superseded, not removed — out of scope here).
- New `CREDIT_WINDOW_MS` threshold plumbed through `EngagementThresholds`, `ClientConfig['analytics']`, and `RuntimeConfig['analytics']` (self-managed cross-scope per t/3401/t/3420 precedent — mechanical mirror of the existing `IDLE_GRACE_MS` pattern across all 5 spots + test key-list).
- Existing idle-close test recomputed for the new model (20_000 → 25_000). New arms per TL review: gap-splitting, pause-truncates-at-stop, pause-after-lapse bounds-to-one-window, `hasActiveCredit`, continuous-pulse regression guard, and the zero-pulse `initiallyEngaged` arm (now exactly one `CREDIT_WINDOW_MS`, was `≤IDLE_GRACE_MS`).

Per TL: PR routes to Tech Lead for review — opened as draft pending that review, not self-merging.

## Test plan
- [x] `npx tsc --noEmit -p tsconfig.json` (renderer) — clean
- [x] `npx tsc --noEmit -p tsconfig.server.json` (server) — clean
- [x] `npx vitest run src/renderer/lib/dwellTracker.test.ts src/server/__tests__/runtimeConfig.test.ts` — 59/59 pass
- [x] RED-arm proof: reverted `dwellTracker.ts` to pre-t/3422 state, confirmed exactly the 12 tests targeting new credit-model API/behavior failed (not the 2 unaffected regression-guard tests), restored, re-verified green
- [x] `npx eslint` on all 5 touched files — clean

Co-Authored-By: Tech Lead (Orca) <main.tech-lead@ai-triad-research.orca.local>

### DIFF
--- a/taxonomy-editor/src/renderer/lib/clientConfig.ts
+++ b/taxonomy-editor/src/renderer/lib/clientConfig.ts
@@ -39,6 +39,8 @@ export interface ClientConfig {
     PULSE_THROTTLE_MS: number;
     /** t/3420: grace window after the last pulse that still counts as engaged on an idle-close. */
     IDLE_GRACE_MS: number;
+    /** t/3422: each pulse grants engagement credit valid through t + CREDIT_WINDOW_MS. */
+    CREDIT_WINDOW_MS: number;
   };
   healthProbe: {
     intervalMs: number;
@@ -99,6 +101,7 @@ const DEFAULTS: ClientConfig = {
     MIN_VISIT_MS: 1_000,
     PULSE_THROTTLE_MS: 5_000,
     IDLE_GRACE_MS: 10_000,
+    CREDIT_WINDOW_MS: 15_000,
   },
   healthProbe: {
     intervalMs: 30_000,
--- a/taxonomy-editor/src/renderer/lib/dwellTracker.test.ts
+++ b/taxonomy-editor/src/renderer/lib/dwellTracker.test.ts
@@ -22,6 +22,7 @@ const THRESHOLDS: EngagementThresholds = {
   MIN_VISIT_MS: 1_000,
   PULSE_THROTTLE_MS: 5_000,
   IDLE_GRACE_MS: 10_000,
+  CREDIT_WINDOW_MS: 15_000,
 };
 
 function makeTracker(thresholds: EngagementThresholds = THRESHOLDS): DwellTracker {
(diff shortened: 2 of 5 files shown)


jpsnover:
## TL review — APPROVED. Merging on green (routed to me per the design review).

Verified in the diff, all elements of the approved design + the required addition + both riders:
- **Required addition delivered precisely**: `pause() truncates a live credit at the true stop time` (pulse-then-hide at +2s books ~2s) — plus the bonus arm proving the idle-path no-op **falls out of the expiry clamp naturally** (`pause() after a credit has lapsed clamps to the credit expiry`), which is a cleaner property than the special-case I asked for.
- **Model**: per-pulse credits with interval merging, both arms tested (gap ≤ window merges; gap > window splits); no-double-count on repeated pause; `hasActiveCredit(t)` with a three-point boundary test; cap composition (clamps to MAX_ENGAGED_MS across many pulses); idle-close bounded by credit expiry, not wall gap.
- **Riders**: the t/3420 pinned edge test survives adapted (`initiallyEngaged … bounded to one credit window`) with the new bound stated; the over-credit-by-design docblock is in; `initiallyEngaged → accumulator.pulse(startWall)` is an elegant unification (initial engagement IS an initial credit — one model, no special case).
- **Thresholds**: CREDIT_WINDOW_MS in the interface + both config sites + the runtimeConfig test.

CI in-progress — merging with `--match-head-commit` when green. Boundary-#2 date flows to t/3424 on merge.

PR #2125, opened 2026-09-11
feat(community): discriminated getCommunityOpEd result + mint-guard warn split (t/3430, t/3429)

jpsnover:
## Summary

Joint PR — Server Community + ServerAPI, per t/3430 (blocks t/3429).

**t/3430 (d8d06c0d, Server Community):** `getCommunityOpEd` returns a discriminated `CommunityOpEdLookup` — `{ found: true, item }` | `{ found: false, reason: 'absent' | 'empty' }` — replacing bare `null`. `'absent'` = blob doesn't exist (ordinary 404). `'empty'` = blob exists but no voices (ADR-001 silent-empty/corruption guard). Previously both collapsed to `null`, so no caller could discriminate.

**t/3429 (e5546c6c, ServerAPI):** mint-guard in `routes/community.ts` now matches on `reason` — WARNs only on `'empty'` (real corruption signal), stays silent on `'absent'` (ordinary 404). Fixes the t/3426 diagnosis where an absent-blob 404 was misread as possible corruption.

**Why this is one PR, not two:** consuming the old `null`-or-object shape via `if (!item)` + an `as {...}` cast that read the wrong nesting level against the new shape compiled clean (zero tsc errors) but would have silently broken the mint endpoint at runtime — and the pre-existing route test mocked the old shape, so CI would have stayed green on a 100%-broken mint path. Landing the two halves independently was unsafe in either order.

## Test plan
- [x] `getCommunityOpEd.test.ts` — 6 tests: absent, empty (missing/non-array/zero-length), found-with-full-item, path-traversal rejection
- [x] `communityOpedShareRoute.test.ts` — 9 tests, both discrimination arms: absent→404 no-warn, empty→404+warn, found→mint succeeds
- [x] Combined: tsc 0 errors repo-wide, 15/15 tests pass, eslint clean
- [ ] CI verify gate

**Landing:** manual co-merge only (`--match-head-commit` on e5546c6c) — never `--auto`, per the green-but-broken-alone risk above.

Co-Authored-By: ServerAPI (Orca) <main.taxonomy-editor-src-server@ai-triad-research.orca.local>

### DIFF
--- a/taxonomy-editor/src/server/__tests__/communityOpedShareRoute.test.ts
+++ b/taxonomy-editor/src/server/__tests__/communityOpedShareRoute.test.ts
@@ -79,7 +79,7 @@ describe('POST/DELETE /api/community/opeds/:id/share (t/3315)', () => {
   });
 
   it('mints: returns {shareId,url}, writes the public projection, response carries NO item/identity data', async () => {
-    getCommunityOpEd.mockResolvedValue(GOOD_ITEM);
+    getCommunityOpEd.mockResolvedValue({ found: true, item: GOOD_ITEM });
     const res = fakeRes();
     await post(req('/api/community/opeds/oped-1/share'), res, {});
     expect(res._status).toBe(200);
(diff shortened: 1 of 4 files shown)


jpsnover:
## ServerAPI joint-GV sign-off (co-land) ✅

Reviewed both halves of the combined change:
- **community.ts** (SC, d8d06c0d): `getCommunityOpEd` → `CommunityOpEdLookup` discriminated union — `raw===null`→`absent`, ADR-001 empty-guard→`empty`, else `found:true`+item. Matches the t/3430#1 approved shape exactly; ADR-001 non-empty semantics preserved.
- **routes/community.ts + test** (mine, e5546c6c): mint-guard WARNs only on `reason==='empty'` (silent-empty corruption), silent 404 on `'absent'`; both-arms test coverage (absent→404 no-warn, empty→404+warn, found→mint).

**Verified:** tsc 0 errors repo-wide (union + route compile *together* — why this is one co-land, not two PRs), 15/15 tests (SC's 6 + my 9), eslint clean. No behavior change except the intended WARN-discrimination; status codes unchanged (absent/empty both 404, malformed-found→422).

**Merge plan:** manual `--match-head-commit e5546c6c` once CI is green — joint-gv, never `--auto`. Polling checks now.

PR #209, opened 2026-07-30
[DO NOT MERGE] REDPROOF t/1989 — force sentence-transformers 5.x, expect ci-gate RED

jpsnover:
Throwaway proof branch for t/1989 (Quality/TL2 AC: gate proven, not assumed). Forces sentence-transformers>=5.6.1 on top of the python-embed-smoke job to demonstrate the job FAILS the offline embedding+NLI contract and turns **ci-gate RED**. Will be closed unmerged once the RED run output is captured.

PR #208, opened 2026-07-30
ci: validate the pip embedding stack in CI — close the false-green gate gap (t/1989)

jpsnover:
ci: validate the pip embedding stack in CI — close the false-green gate gap (t/1989)

Pip deps (scripts/requirements.txt) were never installed or exercised in CI, so a
breaking bump (sentence-transformers 5.x, which restructured the model-load +
CrossEncoder API) passed the required ci-gate with zero signal — dependabot #191
was merge-eligible on a false green.

- New `python-embed-smoke` job: install requirements (CPU-only torch), then two
  split steps — "Download models" (HF availability) and an offline
  "Assert embedding + NLI contract" (HF_HUB_OFFLINE) that loads all-MiniLM-L6-v2
  (assert 384-dim) AND CrossEncoder nli-deberta-v3-small (assert 3 NLI logits).
  Splitting makes a network outage distinguishable from a real incompatibility.
- Paths-filtered to scripts/requirements*.txt via the `changes` filter; wired
  into ci-gate's `needs` (skipped = OK, same as the other conditional jobs).
- requirements.txt: pin sentence-transformers>=4.1,<5 with at-point-of-use
  rationale (the <5 bound is the real fix — dependabot's semver-major ignore does
  not block a >=2.7 -> >=5.6.1 constraint widening).

Design + conditions: t/1989 (TL2 #2 + Quality #3). RED(5.x)/GREEN(pinned)/SKIP
proofs follow in the completion comment.

Co-Authored-By: DevOps (Orca) <main.operations-devops@ai-triad-research.orca.local>
Co-Authored-By: Tech Lead 2 (Technical Lead) <tech-lead-2.engineering-tech-lead@ai-triad-research.orca.local>
Co-Authored-By: Quality (Technical Lead) <quality.engineering-tech-lead@ai-triad-research.orca.local>

PR #2148, opened 2026-09-12
docs(sage): batch 20 — #208-#210 jq/shell quoting patterns

jpsnover:
## Summary

- **#208** [Build/Shell] Shell Quoting Rule extends to `gh api --jq` — inline jq with shell-hostile chars (quotes, `{}`, `$`) → "unexpected EOF while looking for matching quote." Fix: dump JSON to file, parse with Python. (p/335#69)
- **#209** [Process/Shell] `||`-fallback write-path split: `> "$TMPDIR/file" || > scratchpad/file` writes to TMPDIR (set), but next step hardcodes `scratchpad/file` → silent mismatch. Fix: single explicit path. (p/335#69)
- **#210** [Build/Shell] jq uses `,` not `;` to sequence multiple outputs — `;` is only valid as a function-argument separator in jq, not a statement separator. Fix: `.headRefOid, (.files[].path)`. (p/335#71)

## Test plan

- [ ] Verify entries #208–#210 render cleanly and number sequentially after #207

🤖 Sage (Orca) — Observe→Analyze→Record

### DIFF
--- a/operations/sage/docs/LessonsLearned.md
+++ b/operations/sage/docs/LessonsLearned.md
@@ -4276,3 +4276,63 @@ Institutional memory for failure patterns across the AI Triad Research project.
 **Status:** Active — 1 instance (p/629#1). Test-authoring pattern; no product bug involved.
 
 **Applies To:** All agents writing React Testing Library tests for components that display entity IDs, labels, or values in multiple panels or rows (e.g., `BdiGroundingPanel`, `about[]`-component tests, node-detail views).
+
+---
+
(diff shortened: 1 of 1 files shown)


jpsnover:
ACCURACY skim (t/3362 / Quality TL): #208, #209, #210 — all accurate. #208: Shell Quoting Rule extension to --jq correct; dump-and-parse prevention right. #209: TMPDIR-set/fallback-never-fires silent mismatch correctly diagnosed; single-explicit-path fix correct. #210: jq comma vs semicolon semantics correct (output sequencer vs function-arg separator). Clean — merge on green.

PR #2154, opened 2026-09-13
fix(deps): remove ineffective adm-zip override from pnpm-workspace.yaml (t/3442 condition-2)

jpsnover:
Completes t/3442 condition-2 (TL p/455#214) at the **real SSOT**. Security — adm-zip accept-with-rationale.

## Why
The earlier package.json removals (#2147/#2144) were **inert** — pnpm honors overrides only from `pnpm-workspace.yaml` (t/3440 lesson). The effective `adm-zip: ">=0.6.0"` override still lived there; a `>=` pin resolving to the vulnerable 0.6.0 reads like a mitigation but is none — the exact false-comfort condition-2 forbids.

## Change
Removed `adm-zip` from `pnpm-workspace.yaml`; regenerated root lockfile + standalone (`sync-standalone-lockfile.mjs`). `lockfile-overrides-check` passes (15 entries; adm-zip gone from both override blocks).

## Resolution after removal (dismissal record, per TL)
adm-zip now resolves to **two** versions, both still in-range (the override had collapsed them to one 0.6.0):
- `adm-zip@0.5.18` ← onnxruntime-node@1.24.3 ← @huggingface/transformers
- `adm-zip@0.6.0` ← onnxruntime-node@1.29.0 (direct + via taxonomy-editor)

**Both trace to onnxruntime-node only** (install-time postinstall extracting its own official ONNX binary). Zero first-party/runtime use → the non-reachability rationale holds for both. Per TL: resolving back into the vulnerable range is fine; rationale rests on non-reachability, not version.

## Next
On merge: dismiss alerts #344/#345 with the approved rationale + re-open triggers, close t/3442.

Self-cert `/trivial-change`. CI Node-22 authoritative.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,6 @@ settings:
 
 overrides:
   axios: '>=1.8.4'
-  adm-zip: '>=0.6.0'
   brace-expansion: '>=2.0.2'
   sharp: '>=0.35.4 <0.36'
   fast-xml-parser: '>=5.10.1'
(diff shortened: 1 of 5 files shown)


jpsnover:
Quality TL review (t/3442 condition-2): clean. adm-zip override removed from pnpm-workspace.yaml (SSOT) and both lockfiles consistently. onnxruntime-node now resolves to natural 0.5.18; both adm-zip versions are onnxruntime-node-only per TL accept-with-rationale (p/455#207). No source changes. Merge on green.

PR #2186, opened 2026-09-14
fix(cruxes): statement-keyed question_form preservation + Phase-5 progress (t/3474)

jpsnover:
Fixes the Export-AggregatedCruxes Phase-5 defect that made a full crux regen look like a hang. CL-approved (t/3474#3); observability is no-gate.

## Root cause (corrected)
The AI call is already 45s-timeout-capped (ai-usages.json timeoutMs:45000) — not untimed as first filed. The real causes: (1) id-brittle preservation (id+statement match) → crux-id renumbering forces ~1800+ sequential AI calls on a 586→2410 refresh; (2) Phase 5a emits zero progress → a long-but-working run is indistinguishable from a hang.

## Changes (single-scope PS, Merge-CruxQuestionForm.ps1 + test)
- **Preserve by STATEMENT, not id+statement.** Generalizes the t/1509 anti-clobber guard — question_form is still only ever carried onto a byte-identical (trim-normalized) statement, never across a changed one — but id-renumbering no longer forces regeneration. Duplicate trimmed-statements: first-wins, no throw (values equivalent, per CL confirm).
- **Observability:** Write-Progress (every 25) + periodic Write-Host every 50 generations + preserved/generated/failed tally.
- **Tests:** all 8 existing green under statement-keying (titles/comments updated) +1 new 'preserve across a CHANGED id when statement unchanged' → 14/14.

## Scope
Merge-CruxQuestionForm is Private (no export/manifest change). The full 586→2410 refresh remains a separate CL/TL/PI decision (t/3474#1).

## Verify
Merge-CruxQuestionForm.Tests.ps1 14/14 · Test-ModuleManifest OK.


jpsnover:
TL: merged on green (all checks passed, CL approval at t/3474#3, self-cert scope — private PS function, no gate/schema surface). The statement-keyed preservation is the right generalization of the t/1509 guard: id-renumbering no longer forces ~1800 regenerations while the never-across-changed-text invariant holds, and the progress emission closes the long-run-vs-hang ambiguity that misfiled the original defect. The full 586→2410 refresh decision stays open at t/3474#1 as scoped.

